Issue #8, opened 2021-05-18
Permissions of ~/.kube/config too open


garloff:
THe file ~/.kube/config on the kind node have permissions 0644.
This is not what I expect from a file containing credentials. Bad example.
(It may or may not be a concrete risk depending on the access situation, as you can argue that access to the kind node should not be granted to folks who are not normally allowed to use the capi. But this file might also be copied over (using scp -p or rsync) and the too permissive setting might be replicated elsewhere where it makes a difference.)

mxmxchere:
closed by #9

PR #9, opened 2021-05-18
Use mktemp to create safe temp file with 0600 perms.

garloff:
This addresses issue #8.


mxmxchere:
looks good to me, checks should pass when new branches split away from the current master

PR #16, opened 2021-08-02
Feat/use capi images

fynluk:


terraform/files/template/upload_capi_image.sh.tmpl
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+KUBERNETES_VERSION="${kubernetes_version}"
+VERSION_CAPI_IMAGE=$(echo $KUBERNETES_VERSION | sed 's/.\{3\}$//')
```
mxmxchere, line 4: to cut away the patch version the following regex might work better `\.\d*$`

mxmxchere (changes requested):

mxmxchere (approved):
lgtm

Issue #18, opened 2021-08-17
kube-proxy in kind-kind ending in Error (CrashloopBackoff)


garloff:
It appears to crash trying to write to /proc/sys/net/netfilter/nf_conntrack_max, which fails (permission denied).

garloff:
It appears that the container does not have the rights to change sysctls ...
Setting the limit to the wanted 131072 manually on the VM helps the kube=proxy to work.
Not sure that it helps the overall failure

garloff:
With this setting, the testcluster creation works again.

curx:
see https://github.com/kubernetes-sigs/kind/issues/2240 more in detail

garloff:
Thanks, Thorsten!
The error message was good enough that I could guess the solution -- I wish all errrors would be so easy to solve!
I guess the workaround is to just add a sysctl.d setting to the image (or inject it before we deploy kind).
Just adding to the deploy.sh won't work, as you'll need root privs ...

curx:
We can update the [cloud-config](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/master/terraform/mgmtcluster.tf#L43)  to add the needed sysctl settings, to after the bootstrap the kind is ready to use.

garloff:
Fixed by PR #20.

PR #19, opened 2021-08-18
Fix/conntrack sysctl

garloff:
This addresses issue #18 .


garloff:
Discard this, I included a git commit that is not supposed to be part of this PR. Will open new clean PR.

PR #21, opened 2021-08-21
Allow env vars to override ENVIRONMENT, USERNAME, OPENSTACK.

garloff:
This allows you to set the correct ENVIRONMENT in your .bashrc and you
don't need to pass ENVIRONMENT=xxx in every make call.

### DIFF
--- a/terraform/Makefile
+++ b/terraform/Makefile
@@ -2,12 +2,12 @@
 
 # set the env we used 
 #ENVIRONMENT = gx-betacloud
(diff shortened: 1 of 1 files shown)


fynluk:
Looks good to me

fynluk:
Maybe you can fix the DCO check, but for me its not mandatory

garloff:
OOps, I always write git commit -s. Except this time, apparently ... sorry. Fixed now.
Thanks for your review!

PR #22, opened 2021-08-21
Add cleanup script.

garloff:
This adds a new target fullclean that calls a simple clean script.
Unlike ospurge, it is careful not to create collateral damage.
This speeds up my development significantly.

Signed-off-by: Kurt Garloff <scs@garloff.de>


garloff:
This has seen significant testing now, merge.

PR #23, opened 2021-08-21
Update to kind-0.11.1. Override mtu manually.

garloff:
This is my attempt to update to latest kind (0.11.1 currently).
The old way to set mtu to 1400 does not work any more.
We just patch the dockerd startup script (systemd file) and manually set it.
Things *appear* to work.

Signed-off-by: Kurt Garloff <scs@garloff.de>


mxmxchere:
from my side this looks fine. optional enhancement: override the systemd unit file via `/etc/systemd/system/docker.service`

garloff:
> from my side this looks fine. optional enhancement: override the systemd unit file via `/etc/systemd/system/docker.service`

Thanks for your review, Malte!
Patching /lib/systemd/system/docker.service is not ideal, indeed ... it's relatively robust in the initial deployment, but an online update of the docker.io package e.g. would overwrite this change. (Does our maintenance concept assume this might happen?)
When I use /etc/systemd/system/docker.service to override the defaults for docker startup, I would basically reproduce the complete unit file and would then miss any upstream changes ...
So here's my thinking:
1. sed 's/dockerd /dockerd --mtu=1400 /' /lib/systemd/sytem/docker.service > /etc/systemd/system/docker.service
2. systemctl daemon-reload
3. systemctl restart docker.service

So this looks to me like the cleanest approach. Comments?

NB: I did not find a good way to tell kind/dockerd about a different MTU in a config file, which is where I would normally expect such settings.

garloff:
> So here's my thinking:
> 
>     1. sed 's/dockerd /dockerd --mtu=1400 /' /lib/systemd/sytem/docker.service > /etc/systemd/system/docker.service
>     2. systemctl daemon-reload
>     3. systemctl restart docker.service
> 
> So this looks to me like the cleanest approach. Comments?
> 
> NB: I did not find a good way to tell kind/dockerd about a different MTU in a config file, which is where I would normally expect such settings.

Actually, `/etc/docker/daemon.json` is still read by dockerd and it errors out if it find `--mtu=` on the command line and an mtu parameter in daemon.json, even if the values are in agreement.
However, neither sets the mtu on the docker0 interface to 1400, we seem to have to do it manually.
We could of course do it via the runcmd from cloud-init, but this is not reboot safe.
Maybe we have to insert a second ExecStart= line after starting up dockerd.

garloff:
So I ended up injecting a systemd unit file to set the MTU of docker0 to 1400.
Somehow the terraform syntax checker does not like it, though real terraform does not seem to object.
Maybe we should avoid multiline b64 content and rather reference a file?

garloff:
This is the "final" version from my perspective:
* While kind 0.8+ does not use dockershim any more to have k8s talk to the CRI, dockerd on kind host is still supposed to be installed and the configuration with daemon.json is still effective.
* The injected MTU setting should be configurable, which we achieve by avoiding the b64 mess but using plaintext instead.
* Obviously plaintext is preferrable over blobs in git managed files anyhow

There is a terraform syntax check failure, unfortunately. Terraform 1.0.x actually works flawlessly with the mgmtnode.tf as provided, BUT the CI check complains on the syntax. Not sure if we need to fix the test? Maybe upgrade the test to terraform 1.0.x as well? (It reports itself as 0.15.0.)
Hints welcome!

garloff:
OK, terraform fmt did the job.

terraform/mgmtcluster.tf
```diff
@@ -45,21 +46,40 @@ final_message: "The system is finally up, after $UPTIME seconds"
...
+      Description=Docker Set MTU to ${var.kind_mtu}
+      After=docker.service
+      Requires=docker.socket
+      
+      [Service]
+      Type=oneshot
+      ExecStart=/bin/ip link set dev docker0 mtu ${var.kind_mtu}
+      
+      [Install]
+      WantedBy=multi-user.target
```
horazont, line 67: This relies on implementation details of docker (i.e. the interface name) and it also seems racey w.r.t. containers starting (and thus using the wrong initial MTU) and this unit file running.

Also, it only uses `After=docker.service` as synchronization and I don't think there's any guarantee that the network has been created when the service is up.

garloff: Jonas, thanks for the review!
Good points.
I will do some extra testing -- my current belief is that this systemd unit can be dropped again, as the /etc/docker/daemon.json is still effective.

garloff: As suspected: The systemd unit file is not required. PR #57 drops it. Review welcome!

PR #25, opened 2021-08-24
Fix/cleanup full

garloff:


garloff:
Nope, too many unrelated commit in this tree, will redo, sigh.

PR #57, opened 2021-09-03
Drop the uneeded systemd unit file for docker0 MTU.

garloff:
The config file /etc/docker/daemon.json is still fully effective
and all that's needed. So stop setting the MTU in the docker0 device,
as it just adds uneeded complexity/cruft and may be fragile as
Jonas rightly pointed out in PR #23.


horazont (approved):
Negative diffstat, instant-approve ;).

Issue #43, opened 2021-08-31
Add make check target: Deploy testcluster, test app, test CNI and CSI.


garloff:
We should separate steps:
* deploy: do it fast, so bootstrapping to get cluster-api working for customers quickly
* create: also create a testcluster
* check: deploy a test application, also validating that CNI and CSI are functional

garloff:
Options:

* Google microservices test app:
  https://github.com/GoogleCloudPlatform/microservices-demo
  kubectl apply -f ./release/kubernetes-manifests.yaml
  It uses a number of containers with ingress (->octavia) and is a relatively real-world web shop.
  No persistent volumes.
* kuard (O'Reilly - kubernetes up and running) examples. Need to add ingress and volume.
* Separate volume test case is always possible ...

garloff:
https://sonobuoy.io/
is the recommended conformance test tool, it seems.

garloff:
It works (takes ~2hrs) and we pass:
```[1]+  Done                    ./sonobuoy run --wait
ubuntu@capi3-mgmtcluster:~ [0]$ 
ubuntu@capi3-mgmtcluster:~ [0]$ ./sonobuoy status
         PLUGIN     STATUS   RESULT   COUNT               PROGRESS
            e2e   complete   passed       1   337/337 (0 failures)
   systemd-logs   complete   passed       4                       

Sonobuoy has completed. Use `sonobuoy retrieve` to get results.
ubuntu@capi3-mgmtcluster:~ [0]$ results=$(./sonobuoy retrieve)
ubuntu@capi3-mgmtcluster:~ [0]$ ./sonobuoy results "$results"
Plugin: e2e
Status: passed
Total: 5770
Passed: 337
Failed: 0
Skipped: 5433

Plugin: systemd-logs
Status: passed
Total: 4
Passed: 4
Failed: 0
Skipped: 0```

garloff:
Addressed by PR #59.

PR #59, opened 2021-09-15
Feat/sonobuoy tests

garloff:
make check and make quickcheck targets created to run sonobuoy.


garloff:
This addresses issue #43.

Issue #155, opened 2022-03-08
rename dns_nameserver to dns_nameservers


garloff:
The capo template setting is OPENSTACK_DNS_NAMESERVERS with an S, so let's keep the terraform's environment setting in sync.

garloff:
This has been addressed with #164.
Note: We should make sure to properly document this in the release notes.

PR #164, opened 2022-03-10
Feat/multi dns

garloff:
Allow multiple DNS servers.
Reflect this in renaming the environment var to dns_namerservers (with a trailing s), also matching the OPENSTACK_DNS_NAMESERVERS from the capo clusterctl template.
Required a bit of terraform trickery ...


garloff:
Addresses #155.

garloff:
I should have mentioned that the default DNS servers have been changed from quad9 to FF MUC.

Issue #173, opened 2022-03-13
nginx-ingress: 10s delay before getting a connection


garloff:
When talking to services behind a nginx-ingress, we typically see a 10s delay before getting a connection.

garloff:
Trouble is that we have a round-robin loadbalancer talking to all three member nodes, while the nginx L7 LB is only running on one node. So we always try two nodes without success.

garloff:
Adding a healthmonitor helps -- this way, the LB knows that only one member is responding.

garloff:
Resolved by #175.

PR #174, opened 2022-03-13
Add create-monitor=true to could.conf to solve LB issues.

garloff:
See issue #173: When the LB tries to talk to all 3 nodes in round robin
mode, it always typically fails twice before finding the one member
where the nginx L7 LB accepts the connections.
Having a (TCP) health monitor makes the LB aware of which members
respond and which ones do not.

This can be achieved by the global setting create-monitor=true
in OCCM's cloud.conf configuration file.
The only question is: Why does it default to false currently?

### DIFF
--- a/terraform/files/template/cloud.conf.tmpl
+++ b/terraform/files/template/cloud.conf.tmpl
@@ -7,3 +7,4 @@ application-credential-secret="${appcredsecret}"
 [LoadBalancer]
 manage-security-groups=true
 use-octavia=true
+create-monitor=true


garloff:
It looks like nginx-ingress could have asked OCCM for a health-monitored pool.
See
https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
---
loadbalancer.openstack.org/enable-health-monitor

Defines whether to create health monitor for the load balancer pool, if not specified, use create-monitor config. The health monitor can be created or deleted dynamically. A health monitor is required for services with externalTrafficPolicy: Local.

Not supported when lb-provider=ovn is configured in openstack-cloud-controller-manager.

garloff:
https://kubernetes.github.io/ingress-nginx/deploy/#cloud-deployments
* We can chose externalTrafficPolicy=Cluster to not require a health monitor
* ... or chose =Local and save one hop and need a health monitor.

garloff:
And we can edit the nginx-ingress deployment yaml:
`
apiVersion: v1
kind: Service
metadata:
  annotations:
    loadbalancer.openstack.org/enable-health-monitor: "true"
    #loadbalancer.openstack.org/proxy-protocol: "true"
  labels:
    helm.sh/chart: ingress-nginx-4.0.2
    app.kubernetes.io/name: ingress-nginx
    app.kubernetes.io/instance: ingress-nginx
    [...]
    app.kubernetes.io/component: controller
  name: ingress-nginx-controller
  namespace: ingress-nginx
spec:
  type: LoadBalancer
  [...]
`

garloff:
Being at it, we could also enable the proxy protocol:
apiVersion: v1
kind: ConfigMap
  labels:
    [...]
  name: ingress-nginx-controller
  namespace: ingress-nginx
data:
  proxy-protocol: "True"
  real-ip-header: "proxy_protcol"
  set-real-ip-from: "0.0.0.0/0"
```
and
```
kind: Service
metadata:
  annotations:
    loadbalancer.openstack.org/enable-health-monitor: "true"
    loadbalancer.openstack.org/health-monitor-timeout: "1"
    loadbalancer.openstack.org/health-monitor-max-retries: "2"
    loadbalancer.openstack.org/proxy-protocol: "true"
  labels:
    [...]
  name: ingress-nginx-controller
  namespace: ingress-nginx
spec:
  type: LoadBalancer
  externalTrafficPolicy: Local
```

garloff:
We prefer the approach taken by #175.

garloff:
We should IMVHO still consider this one - not everyone uses the nginx-ingress as deployed with our clusters but may use nginx deployment files from somewhere else. They will have a higher chance to work well with a health monitor enabled default...

PR #175, opened 2022-03-15
Use kustomize to enable monitor (and proxy).

garloff:
This is a different approach to fixing #173:
We can explicitly ask the OCCM to use a health monitor, see discussion
in #174. This is preferable, as we can leave the LB defaults as is.

Use kustomize to actually do the changes to the downloaded nginx
deployment file, as this is a rather robust way.

Being at it, also allow enabling the proxy protocol, so out backends
see the real IP (in the http header set by our nginx) and not just the
VIP from the LB.

We save the generated nginx deployment file, so we can use it for
cleanup in delete_cluster.sh again.

Also: Ship the nginx-i-c-v1.1.2 deployment file directly.


garloff:
Notes to myself:
* Improve error handling: Error out if kustomize fails.
* Test proxy protocol

batistein (approved):
I have not tested it yet, however, kustomize and ingress-nginx conf looks right to me.

garloff:
OK, we correctly get both the X-Forwarded-For: and X-Real-IP: headers passed via this construct if `NGINX_INGRESS_PROXY: true` is set.

Issue #117, opened 2022-01-26
Split capi mgmt node creation from workload cluster management.


garloff:
This builds on top of #109 (and also touches #107):
Currently we conflate the creation of the capi management node with the creation/management of a testcluster (which has currently the constant name testcluster, see #107).
Both are configured in one environment file, which is unfortunate. More precisely, there are four kind of settings:
(A1) There are cloud-specific settings for the mgmt node.
(A2) Some version settings (which capi version, other tools such as k9s ...) for the mgmt node
(B1) The workload clusters also require a few cloud specific settings (flavor, AZ)
(B2) ... and some settings that describe the k8s workload cluster (k8s version, preinstalled OCCM, CSI, metrics, .... no of nodes ...)

This should be properly split.

While it is useful (e.g. for CI purposes) to have the ability to create a testcluster and run tests (e.g. sonobuoy) when creating the capi management node, it is not the typical use scenario. Instead, we should have a gitops style pipeline, where the wanted workload clusters are created and configured based on git repos. Ideally, we make the test cluster one application of this approach.
We may keep a way to test and prototype things using the management scripts and continue to enhance them a bit further for this.
The assumption however currently is that we'll typically use the helm based workload clusters ...

Issue #160, opened 2022-03-08
Create ability to put clusters in own projects


garloff:
Extending #109 we could also have the option to create a new project for a cluster.
Flow would be
- Create project
- Create app-creds for projects
- Share image with project
- use the app-creds ...
All OpenStack commands should reference this new project, except image registration/sharing.

On cluster deletion, we would then also delete the project again, after ensuring that nothing is left in there.

garloff:
Some development for this has happened in the branch feat/per-cluster-appcred
https://github.com/SovereignCloudStack/k8s-cluster-api-provider/tree/feat/per-cluster-appcred

garloff:
Mostly done.
Missing:
* Documentation (how to add 2nd/3rd/nth cloud in clouds.yaml and then reference it in clusterctl.yaml)
* Add some smartness to avoid redownloading and reregistering images, as we can just share between projects (that we control)

Issue #109, opened 2022-01-06
Use unrestricted app cred for capi mgmt node and 2ndary restricted app creds for capo and for CSI and OCCM


garloff:
We should have capability to create further app creds for clusters on the capi mgmt node.
This is a first step into managing clusters across clouds from the same mgmt node.

We should then create two (!) restricted app creds per cluster:
- One is for capo: This one can not be accessed / abused from within the workload cluster.
- One if for CSI and OCCM, allowing create OpenStack resources (loadbalancers, volumes, ...) from within the workload cluster

The latter is subject to abuse if hostile users within the cluster want to do evil things. So be prepared to independently revoke these app creds.

garloff:
This should also prepare for having multiple clouds configured in ~/.config/openstack/clouds.yaml and secrets.yaml and avoid assuming particular formatting of it. (It could as well have std user/password combo in there.) 

This also nicely decouples the need for OCCM clouds.yaml to have project_id (despite using appcreds, where std openstackclient does neither need nor tolerate project_id).

garloff:
Work on deployment node done (we generate an unrestricted application credential, #177),
but not yet the creation of per-cluster app-creds.

garloff:
This is fully addressed in PR #226.
Once the PR is merged, we can close this one.

chess-knight:
@garloff can you close this one, please?

PR #176, opened 2022-03-17
Feat/repo on mgmtcluster

garloff:
This PR changes the way we work on the management node.
By having the git repo checked out on the management node, we can do incremental updates from upstream without having to redeploy the management node (and thus destroy all the workload clusters).
This was the last major outstanding feature for R2.


garloff:
This fixes #117.
It also prepares #109 and #160 by having separate sub directories on the management node for each cluster.

Issue #184, opened 2022-03-18
new proxy protocol breaks pod to ingress communication


flyersa:
the new proxy protocol setting in the nginx ingress breaks pod to ingress lb communication. e.G pods cannot request from ingress setups deployed in the cluster in the same network, it will only work from external. 

values in ngrinx-ingress.yaml:

loadbalancer.openstack.org/proxy-protocol: "true"
use-proxy-protocol: "true"

This also breaks cert-manager with letsencrypt cluster issuer since the cert-manager is not able to curl for the .acme challenges. In proxymode connections only work from outside the network, with proxy-protocol active pods running in the cluster cannot access ingress urls (empty response).

setting proxy-protocol to false ín the current setup will make the ingress only available from inside the cluster, but then the ingress will stop working from outside.

digitalocean on there own provider face something similar and there is a workaround by setting some kind of hostname to the ip of the deployed loadbalancer (step 5 here https://www.digitalocean.com/community/questions/how-to-support-internal-traffic-with-proxy-protocol-enabled-on-a-kubernetes-loadbalancer) .  similar reference here: https://github.com/kubernetes/cloud-provider-openstack/issues/1287

garloff:
Unless someone can debug this quickly and determine a clean solution, my spontaneous approach would be to change the default here to disabled and document the potential trouble. This is implemented in #185.

garloff:
Jonas, Sven, Fynn, if you happen to have something like this before and know what the clean solution (TM) is, please let us know.
Otherwise, we'll just do the disablement in the default settings for R2 and then investigate in more detail afterwards. THANKS!

horazont:
I don't think there's an easy solution, because you'd have to configure the nginx ingress to except the proxy protocol only for the LB source IPs.

Dynamically accepting the proxy protocol could be a security issue, so it needs to be a hard on/off switch (just like with X-Forwarded-For).

Disabling it seems like the sanest choice.

garloff:
As we have a safe default now, this is now an enhancement:
Figure out if there is a way to enable proxy protocol without breaking internal connections.
If we don't find one, we may close this ...

garloff:
See script `nginx_proxy_realip.sh` from PR #199 to test this.

garloff:
Fixed by #323

PR #185, opened 2022-03-18
Disable the usage of proxy protocol by default.

garloff:
If nginx expects inbound connections to use it, this works fine when the
connections come from the LB that has been configured to use the proxy
protocol. However, talking to ingress directly from the pods does then
no longer work (as noone injects the proxy proto then).

This means we should probably not enable it by default.

This should address #184.


garloff:
Let's revert this for now.

PR #202, opened 2022-03-30
Register cinder storage class.

garloff:
This got accidentially dropped when we started to have version dependent
downloads for the cinder driver.

We also had some cinder pieces that actually failed to be version
dependent, this is fixed now as well.


garloff:
Tested successfully on k8s-1.19 -- 1.23.
It fixes #203 successfully.

PR #217, opened 2022-06-02
Improve attaching/detaching NICS

garloff:
add_cluster-network.sh: Save unnecessary server list calls:
  We know the name of the -mgmtcluster host.
remove_cluster-network.sh: Determine need to delete NIC and
  determine net and NIC name. Exit if nothing needs to be done.
delete_cluster.sh: Call remove_cluster-network.sh, just in case
  some cleanup is needed. It's harmless if not. If it is, it helps
  us to ensure that kubectl delete cluster can succeed.

### DIFF
--- a/terraform/files/bin/add_cluster-network.sh
+++ b/terraform/files/bin/add_cluster-network.sh
@@ -22,7 +22,9 @@ findnewnic()
 	return 1
 }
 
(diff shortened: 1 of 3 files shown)


garloff:
This is 3.x material, not necessarily 3.0.x.

PR #221, opened 2022-06-14
:bug: fixes kubernetes versions

batistein:
Removes a comma in the list of supported k8s versions.
Bug was found while debugging the occm which couldn't be deployed on v1.23.1, because of a wrong constructed path due to the comma.


garloff (approved):
Thanks for the fix!

PR #226, opened 2022-07-03
Feat/per cluster appcred

garloff:
This is the long awaited capability to have
(a) per Cluster Application Credentials
(b) the capability to manage clusters from one mgmt node/cluster in several clouds/projects

In order to achieve this, we stop assuming that ~/.config/openstack/clouds.yaml (and secure.yaml) may only contain one cloud and needs to necessarily be configured with an appcred (although this remains the standard for the mgmt host deployed by terraform in k8s-cluster-api-provider). Instead we have code that extracts just the pieces that we need, see print-cloud.py.
With this, we are now ready to create an application credential per cluster. We will append the config to the clouds.yaml file and also generate cloud.conf and the b64 encoded clouds.yaml from this cloud auth setting.

create_cluster.sh -- or more precisely the create_appcred.sh called from there is capable to detect if we already have an appcred or if the cluster is running without one (old clusters may coexist with new ones). We do tweak the OPENSTACK_CLOUD setting in clusterctl.yaml, so everything is using the new app cred.

On delete_cluster.sh, we remove the app cred cloud auth config from clouds.yaml again, delete the appcred and reset the OPENSTACK_CLOUD in the clusterctl.yaml that is left behind.


garloff:
This addresses issue #109.

It also is 90% of what's needed for #160.
The missing 10%:
* We could do some clever image sharing here in order to save image storage capacity and time.
* Document how to configure a second/third/nth project in ~/.config/openstack/clouds.yaml and how to reference it in clusterctl.yaml when creating a new cluster. (Yes, it is straightforward :-) )

berendt:
"We could do some clever image sharing here in order to save image storage capacity and time."

Why not let the CSP provide the images as public and then use them? I think it is not necessary to upload the images themselves.

flyersa:
I would also say that in the current state it is a good idea to keep the image download.  It helps alot with testing

batistein (approved):

garloff:
Release notes should be written.

garloff:
Upgrade documentation missing, otherwise completed.

PR #233, opened 2022-08-11
Feat/sign certs

garloff:
Helper script for signing CSRs with a custom CA (e.g. the one from kubernetes).
This should not be needed, as `kubectl certificate approve <CSR>` should do have the built-in signers do the job.


garloff:
We now know how to check the validity of the cert signing requests and have a mode where we can run this.
It is a PoC, as mentioned in the header. Nevertheless, I think it would be useful to merge and have in the toolbox.

PR #241, opened 2022-08-16
Fix "remove_cluster-network.sh: command not found"

### LINKED ISSUE #240: /home/ubuntu/bin/delete_cluster.sh: line 40: remove_cluster-network.sh: command not found
When I clean an environment with make clean the following error occurs:

```
[...]
validatingwebhookconfiguration.admissionregistration.k8s.io "ingress-nginx-admission" deleted
No resources found in default namespace.
/home/ubuntu/bin/delete_cluster.sh: line 40: remove_cluster-network.sh: command not found
cluster.cluster.x-k8s.io "testcluster" deleted
[...]
```

berendt:
Closes #240

Signed-off-by: Christian Berendt <berendt@23technologies.cloud>

### DIFF
--- a/terraform/files/bin/delete_cluster.sh
+++ b/terraform/files/bin/delete_cluster.sh
@@ -37,7 +37,7 @@ if grep '^ *OPENSTACK_ANTI_AFFINITY: true' $CCCFG >/dev/null 2>&1; then
 	fi
 fi
 # Detach network interface (if ever attached)
-remove_cluster-network.sh "$CLUSTER_NAME" >/dev/null || true
+~/bin/remove_cluster-network.sh "$CLUSTER_NAME" >/dev/null || true
 # Tell capi to clean up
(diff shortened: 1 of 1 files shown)


garloff (commented):
This should not be needed. bootstrap.sh does inject a PATH=$PATH:~/bin into ~/.bash_aliases, which is read by the default ~/.bashrc if it exists.
I assume it is needed as I can see from issue #240.
It seems the ssh command from make does call bash in a way that ~/.bashrc is not sourced?
It would be safer to add a PATH=$PATH:~/bin at the beginning of delete_cluster.sh with an appropriate comment of why it may be needed. This is more robust.
For now, ignoring the error is safe unless you have manually called add_cluster_network.sh [CLUSTERNAME], which is only ever needed for debugging.

garloff:
Superceded by #247.

Issue #243, opened 2022-08-16
Error while removing application credential in bin/delete_cluster.sh


berendt:
```
Removing application credential capi-testcluster-appcred ...
./bin/delete_cluster.sh: line 74: /home/ubuntu/tmp/clouds-no-capi-testcluster.yaml: No such file or directory
```

garloff:
Same reason as #240: Missing path causes print-cloud.py not to be found.

garloff:
Fixed by #247.

PR #247, opened 2022-08-19
Source ~/.bash_aliases from some scripts.

garloff:
We need to do this when these scripts are invoked remotely (via ssh from
the Makefile), as non-interactive shells do not read ~/.bashrc.

Do this for cleanup.sh, deploy_cluster_api.sh and sonobuoy.sh which are
the only scripts called directly from the Makefile AFAICS.

This should address issue #240.


garloff:
Closes #240 and closes #243.

garloff:
Tested successfully.

PR #257, opened 2022-08-23
Use a more robust method to grab latest cilium.

garloff:

### DIFF
--- a/terraform/files/bin/clusterctl_template.sh
+++ b/terraform/files/bin/clusterctl_template.sh
@@ -32,7 +32,7 @@ CLOUD_CONF_ENC=$(base64 -w 0 ~/$CLUSTER_NAME/cloud.conf)
 #Get CA and Encode CA
 # Update OPENSTACK_CLOUD
 if test "$OS_CLOUD" != "$OLD_OS_CLOUD"; then
-  echo "#Info: Changing OPENSTACK_CLOUD frpm $OLD_OS_CLOUD to $OS_CLOUD"
+  echo "#Info: Changing OPENSTACK_CLOUD from $OLD_OS_CLOUD to $OS_CLOUD"
   yq eval '.OPENSTACK_CLOUD = "'"$OS_CLOUD"'"' -i ~/$CLUSTER_NAME/clusterctl.yaml
   sed -i "/^OPENSTACK_CLOUD:/a\
 OLD_OPENSTACK_CLOUD: $OLD_OS_CLOUD" ~/$CLUSTER_NAME/clusterctl.yaml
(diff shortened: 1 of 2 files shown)


terraform/files/bin/clusterctl_template.sh
```diff
@@ -32,7 +32,7 @@ CLOUD_CONF_ENC=$(base64 -w 0 ~/$CLUSTER_NAME/cloud.conf)
 #Get CA and Encode CA
 # Update OPENSTACK_CLOUD
 if test "$OS_CLOUD" != "$OLD_OS_CLOUD"; then
-  echo "#Info: Changing OPENSTACK_CLOUD frpm $OLD_OS_CLOUD to $OS_CLOUD"
+  echo "#Info: Changing OPENSTACK_CLOUD from $OLD_OS_CLOUD to $OS_CLOUD"
```
joshmue, line 35: This change does not seem to be related to the actual change.

(Nothing critical of course, but I was assigned to review, so I review :D)

terraform/files/bin/deploy_cluster_api.sh
```diff
@@ -48,7 +48,8 @@ sudo mv calicoctl /usr/local/bin
 
 # Install cilium
 # TODO: Check signature
```
joshmue, line 50: Kind of unrelated: This TODO can be removed, right?

joshmue (approved):
LGTM generally.

batistein (approved):

terraform/files/bin/deploy_cluster_api.sh
garloff: Well downloading the binary and the cryptographic hash from the same place and checking that they match does not add any security against attackers that want to inject evil code into your deployment ... as both code and checksum come from the same source at the same time and are thus prone to the same potential attack.
It helps against corruption during transmission.
We would need to check cryptographic signatures (with a list of trusted keys that we retrieve beforehand and keep in the repo) or at least have a collection of known good cryptographic checksums stored in the repo and compare the downloaded binary ...

terraform/files/bin/clusterctl_template.sh
garloff: You are right -- I was lazy and did the *cosmetic* fix here, while it should have been done in a separate branch.
Good that you spot this -- it makes me more disciplined in the future.

terraform/files/bin/deploy_cluster_api.sh
joshmue: Right. Made me think whether it'd be a good idea to create a GH Issue/Epic to decide on preferences regarding tooling/concepts providing such "Supply Chain Security" (like sigstore, TUF...), but that does not seem necessary, as there is a tender planned for lot 9b already.

Issue #242, opened 2022-08-16
LICENSE file missing


berendt:
There is no LICENSE file in this repository. Should be added.

garloff:
Mixture of ASL-2.0 and CC-BY-SA-4.0 currently, resulting in complete works being CC-BY-SA-4.0.

garloff:
Should try to relicense under ASL-2.0 to have a clean picture.
AI Kurt to track down contributors that have contributed under CC-BY-SA-4.0.

garloff:
This has been addressed with PR #261.

PR #261, opened 2022-08-26
Add LICENSE to address issue #242.

garloff:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,508 @@
+# Sovereign Cloud Stack Kubernetes Cluster API Integration license information
+
+## Overview 
+
+The code in this repository consists of pieces licensed under the
+Apache software license v2.0 (ASL2) and under 
+Creative-Commons v4 with attribution and share-alike clauses (CC-BY-SA-4.0). 
+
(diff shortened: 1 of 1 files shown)


garloff:
Note:
* The repository contains pieces under Apache-2.0 and und CC-BY-SA-4.0, which are both fine licenses.
* These can be combined with the result that the combined work is under CC-BY-SA-4.0.
I just stated this ... if we believe this is too confusing, we could start a debate with the contributors about relicensing.

itrich (commented):
I'm not absolutely sure whether [Licensee](https://github.com/licensee/licensee) will parse this as valid `LICENSE` file. Moreover, I'm concerned about how tools like [ORT](https://oss-review-toolkit.org/) will handle this license identifier.

garloff:
OK, so the plan of record is to go through the various scripts, and ensure we talk to all authors of CC-BY-SA-4.0 pieces and ask them to agree to Apache-2.0.

Vad1mo (approved):

garloff:
Everything converted to Apache-2.0 now.
Please review!

Issue #269, opened 2022-09-12
git repo choice: Need to translate git@HOST: to https://HOST/


garloff:
When automating the git repository choice, we need to avoid using accounts with login, as they will not typically work on a freshly deployed management cluster host.
See https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/266#issuecomment-1243485647

garloff:
Fixed by #270.

PR #266, opened 2022-09-07
Add possibility to choose git repo

fynluk:
Implements #237 

Signed-off-by: Fynn Späker <spaeker@23technologies.cloud>

PR #270, opened 2022-09-12
Address #269: Avoid using git accounts.

garloff:
When making the git repo configurable that the automation in the cluster management hosts comes from (#266), we detected the existing git repo and pass it down from the Makefile.

However, for developers, there is a significant likelihood that they are using a clone checked out with their account git@github.com. This will not work on the management host, as the account is not deployed there.

So translate this to an anonymous https:// URL before passing down.


fynluk:
Yes, you are completely right. 
Thx for the fix, regex looks good to me

garloff:
Thanks for fast feedback -- if you formally approve the PR, I don't need to override the branch protection (though I would do so based on your factual review, if need be).

fynluk (approved):
/lgtm

PR #276, opened 2022-09-17
Upgrade to nginx_ingress controller v.1.3.0

garloff:
This marks the transition to do leader elections with leases. Upgrade from earlier versions should go to v1.3.0 first and then onwards.


batistein (approved):

Vad1mo (approved):

garloff:
Thanks for your reviews, Sven and Vadim!

PR #278, opened 2022-09-17
Make image metadata safe against Feb 29 dates.

garloff:
It's not a strict requirement as it's a private image, but providers might use this to register and provide public images, so better have all properties.


garloff:
Oh, double commit ...
So the changelog for this one should become:
Make image metadata safe against Feb 29, as that's the only incremental change.

garloff (commented):
Does still look and test correctly on second view.

PR #279, opened 2022-09-17
Enable etcd tweaks by default.

garloff:
This keeps the etcd clusters stable even in somewhat awkward situations with network storage (or high steal time, though that does not seem to be as muh of a problem).

It's better to build stable 3/5/7 node clusters for the control plane and risk losing etcd updates when a node really crashes than not having stable clusters.

So let's default those tweak to be enabled.


garloff:
To support the reviewers:
Here's what `ETCD_PRIO_BOOST: true` does:
`if test "${ETCD_PRIO_BOOST}" = "true"; then grep 'hearbeat\-interval' /etc/kubernetes/manifests/etcd.yaml >/dev/null || sed -i '/\-\-trusted\-ca\-file/a\    - --heartbeat-interval=250\n    - --election-timeout=2500' /etc/kubernetes/manifests/etcd.yaml; sed -i 's@\(cpu.\) 100m@\1 300m@' /etc/kubernetes/manifests/etcd.yaml; fi`
and here's what `ETC_UNSAFE_FS` does:
`if test "${ETCD_UNSAFE_FS}" = "true"; then mount -o remount,barrier=0,commit=20 /; sed -i 's@errors=remount-ro@errors=remount-ro,barrier=0,commit=20@' /etc/fstab; fi`

batistein:
@garloff I would think about shifting from modifying the generated manifest with sed to directly editing the KubeadmControlPlane object under extraArgs, that would be a more neater approach.

garloff:
> @garloff I would think about shifting from modifying the generated manifest with sed to directly editing the KubeadmControlPlane object under extraArgs, that would be a more neater approach.

The PrioBoot pieces you mean?
Something like in configmap kube-system/kubeadm-config:
```
data:
  ClusterConfiguration: |
[...]
    etcd:
      local:
        dataDir: /var/lib/etcd
        extraArgs:
          heartbeat-interval: 250
          election-timeout: 2500
```
You probably have a better idea than using kustomize (or the even more cumbersome k8s patching mechanism) on the ClusterConfiguration?
will appending a 
```
--- 
apiVersion: kubeadm.k8s.io/v1beta3
kind: ClusterConfiguration
etcd:
  local:
    dataDir: /var/lib/etcd
    extraArgs:
      heartbeat-interval: 250
      election-timeout: 2500
```
to the cluster-template just do the trick?

batistein:
I would template the cluster-template before applying (could be done via kustomize or a helm chart)

This is how we (Syself) do it for KubeadmControlPlane (kcp):
Here a short snippet for the etcd part. 
input from CI test: https://github.com/syself/charts/blob/main/charts/capi-cr/ci/hetzner-cluster-template.yaml#L147-L154
templating logic: https://github.com/syself/charts/blob/main/charts/capi-cr/templates/controlplane.cluster.x-k8s.io/kubeadm_control_plane.yaml#L80-L85

garloff:
The priority boosting is included in #282.
We don't have agreement that `ETCD_UNSAFE_FS` defaulting to `true` is a good idea:
* There are legit reasons to run single-controller k8s clusters
* Even for multi-controller k8s clusters, there is an option to only run one etcd (well, if it's on safe ceph storage) -- this is not supported out of the box with our capi automation though
* And for real multi-controller k8s clusters, having local SSD/NVMe storage is of course still the best option

So we reject this change for now, only adding some better documentation.

PR #291, opened 2022-09-20
Remove hint to CoC in R3 release notes

itrich:
Should we really give that hint to the CoC in the release notes? I think this may sound a bit daunting – especially if we end the sentence by "failing to meet the expectations".


garloff (approved):
Agree, we don't need to reiterate the CoC.

PR #293, opened 2022-09-20
Feat/upgrade guide

garloff:


garloff:
See https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/feat/Upgrade-Guide/doc/Upgrade-Guide.md

batistein (approved):
super detailed and everything so far in it for a regular update. Only thing I wonder is how to update when changes have been made in the cluster templates. Due to the fact that so far one could not set any extra machine deployments, this use-case has occurred more often.

garloff:
Thanks for your review, @batistein !
We should create more advanced cluster templates in the future to accommodate more complex use cases.

The Upgrade Guide is incomplete -- we could well add a section that explains what happens if you have to modify the cluster-template.yaml.

Issue #298, opened 2022-09-26
upload_capi.sh has lost its eXecute permission


garloff:
```
Waiting for image ubuntu-capi-image-v1.24.4 to become active: Image ubuntu-capi-image-v1.24.4 does not exist, create ...
/home/ubuntu/bin/wait_capi_image.sh: line 15: /home/ubuntu/bin/upload_capi_image.sh: Permission denied
```
'nuff said.

garloff:
Fixed in main and the two maintained/ branches.

PR #299, opened 2022-09-26
eXecutable bit had been lost on upload_capi.sh (#298).

garloff:
Reestablish it.
Maybe in the future, we need to add a workaround for this, as this might happen whenever people touch files without permission-aware Operating Systems / File Systems.


garloff (commented):
No brainer.

PR #301, opened 2022-10-18
Feat/k8s latest 1.25.3

garloff:
Upgrading k8s to latest versions.
Upgrading OCCM also to latest version, but limit patchlevel on v1.25.x to be no larger than k8s patchlevel.

Candidate for maintained/v4.x.0 and maintained/v4.0.x branches.

### DIFF
--- a/terraform/files/bin/create_cluster.sh
+++ b/terraform/files/bin/create_cluster.sh
@@ -76,6 +76,12 @@ if test "$CONTROL_PLANE_MACHINE_COUNT" -gt 0 &&  grep '^ *OPENSTACK_ANTI_AFFINIT
 	fi
 fi
 
+# Patch registry location for k8s >= 1.25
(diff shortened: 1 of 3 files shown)


Vad1mo (approved):

garloff:
Thanks for the review, Vadim!
In testing, I found that we need to replace k8s.gcr.io with registry.k8s.io for 1.25.x to work.
(The coredns service in the latest versions will not be served via k8s.gcr.io, it seems, see #303.)
So I had to push another change in here.

garloff:
One would expect that registry.k8s.io works for all k8s versions, but that vastly overestimated the considerations that that community has for users of older versions.
We will need code that patches cluster-template depending on the k8s version, sigh.

batistein:
Is this backported to R3? I really would like to use it for the next integration cluster

batistein (approved):

garloff:
Thanks for the review, Sven.
I think that all three PRs merged today should go into both `maintained/v4.x` and also `maintained/v4.0.x`.

PR #304, opened 2022-10-18
calico had not really been updated to v3.24.1 before.

garloff:
In commit 176cc01b3b509bab6bf83cae3fcf20730d8458da, we wanted to update calico to v3.24.1, but we really only had succeeded in updating the calicoctl binary. We had overlooked that the deployment yaml was stored in kubernetes-manifests.d/ and still old (v3.22.1).

Unfortunately, the v3.22.1 does not work with k8s-1.25, causing serious failure.

Note: We currently have a parameter that allows us to set the calico version from the terraform variables, which is mostly ineffective, as it only affects the calicoctl binary. Will report and fix this in a subsequent issue.

### DIFF
--- a/terraform/files/kubernetes-manifests.d/calico.yaml
+++ b/terraform/files/kubernetes-manifests.d/calico.yaml
@@ -4180,7 +4180,7 @@ rules:
     resources:
       - endpointslices
     verbs:
-      - watch 
+      - watch
       - list
   - apiGroups: [""]
     resources:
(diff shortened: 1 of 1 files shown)


garloff:
I tested that this does not break old 1.21.x k8s. (I have not tested any older versions.)
And for 1.25, we need it. Review would be appreciated.

batistein (approved):

PR #306, opened 2022-10-19
Fix typo SCS-2D -> SCS-2C as flavor default for controller.

garloff:
Note: This gives us SCS-2C:4:20s, which is a flavor with 2 dedicated cores and a (local) 20GB SSD disk. This would be a good choice for control plane nodes and what was intended before. We have plans to standardize a SCS-2V:4:20s flavor, so this will eventually become the default here. We have not observed issues with excessive steal time for etcd, so this is believed to be good enough.


batistein (approved):

batistein:
Thank you very much happy to test it out tomorrow!

PR #309, opened 2022-11-29
Add latest k8s patches.

garloff:
v1.22.15 -> .16
v1.23.13 -> .14
v1.24.7  -> .8
v1.25.3  -> .4

(No new OCCM versions published.)

This change makes the latest versions known to the scripts, which can use the v1.<yy>.x syntax -- with <yy> being a number, e.g. 25 and x being the letter x which does get translated to latest according to this table.


tibeer (approved):
Not tested, yet looks very valid!

PR #310, opened 2022-11-29
Update default version to capi-v1.2.6, capo-v0.6.4.

garloff:
Update to capo version v0.6.4:
* Implement conversion for OpenStackClusterTemplate CRD
* Fixed logger arguments

Update default capi version to v1.2.6:
* ClusterCacheTracker: Per cluster locking avoids unreachable clusters to block others
* kubekey provider
* finalizer reconciliation for TOpology MachineSets and MachineDeployments

Update default capi version to v1.2.5:
* Use registry.k8s.io images (rather than k8s.gcr.io)
* Avoid clusterctl move to change managedFields
* Run kubeadmconfig reconciliation after datasecretname setting
* go1.18.8 and bump dependencies to address CVEs.


tibeer (approved):
Not tested, but looks good since there are only version bumps.

garloff:
Thanks, Tim!

PR #314, opened 2023-01-18
Pass cluster name when waiting for machine ready.

garloff:
Otherwise, if we manage multiple clusters, we will wait for other cluster's control plane machines being ready -- which might cause hangs/timeouts, in case multiple are started in an interleaved fashion (or one cluster hangs).


matofeder (approved):
LGTM. Modified commands have been tested in the local CAPI environment.

garloff:
THANKS, Matej!

PR #315, opened 2023-01-18
Feat/latest k8s 202301

garloff:


batistein (commented):
kubernetes version v1.26 is not supported see: https://github.com/kubernetes-sigs/cluster-api-provider-openstack#compatibility-with-cluster-api-and-kubernetes-versions

https://github.com/kubernetes-sigs/cluster-api/issues/7671

I would therefore wait before adding the newest k8s version.

garloff:
I could exclude v1.26.0 from the patch indeed.

Or add a note to the documentation that it needs more validation...
I have done some basic testing of 1.26 and it just worked. I would next throw the CNCF cert suite at it (with sonobuoy). I would also add in a few capi specific things like testing rolling upgrades and clusterctl moves. If all succeed, we may get away with just a warning.

garloff:
CNCF tests pass with v.1.26.0.
```
2023-01-20T10:29:48Z                                                                                                                                      
         PLUGIN     STATUS   RESULT   COUNT                 PROGRESS                                                                                      
            e2e   complete   passed       1   Passed:366, Failed:  0                                                                                      
   systemd-logs   complete   passed       6                                                                                                               
                                                                                                                                                          
Sonobuoy has completed. Use `sonobuoy retrieve` to get results.                                                                                           
=== Collecting results ===                                                                                                                                
INFO[0000] delete request issued                         kind=namespace namespace=sonobuoy                                                                
INFO[0000] delete request issued                         kind=clusterrolebindings                                                                         
INFO[0000] delete request issued                         kind=clusterroles                                                                                
Plugin: e2e                                                                                                                                               
Status: passed                                                                                                                                            
Total: 7072
Passed: 369
Failed: 0
Skipped: 6703

Plugin: systemd-logs
Status: passed
Total: 6
Passed: 6
Failed: 0
Skipped: 0
```
which is not so surprising. Once capo has setup the cluster, it is pretty much out of the loop and OCCM, CSI, CNI are the ones that could spoil the results. These are on v.1.26.0 already.

garloff:
I successfully did a rolling upgrade (both control plane and workers).
I successfully moved the cluster-api objects with clusterctl move into the workload cluster and back into a newer version of kind.
I checked health of services (e.g. the running nginx) to ensure things work.
I could not see anything fail.

So I would proceed as follows:
* Leave v1.26.0 active. (I would of course NOT make it default, but users can chose it.)
* I would output a warning that v1.26.x is not officially supported yet, as it needs more upstream testing.
Makes sense?

garloff:
According to the team discussion this morning, we should enable v1.26.x, but mark it as Technical Preview until capi/capo have picked up official support for it.
I'll double check how we can best make this visible, the messages added in https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/315/commits/93087c451a473dcf71bd8e8c57f8407f77f43c66 may not be visible enough.
I'll ask for a review again after checking.

garloff:
OK, we now require `--allow-preview-versions` in order to support v1.26+.
Please review @batistein, @joshmue, @jschoone, @chess-knight

chess-knight (approved):
I tested it with v1.26.0 and it works.

garloff:
Thanks for testing!

PR #316, opened 2023-01-18
Feat/use vim

garloff:
Admittedly, this may not please everyone, but a certain amount of opinionatedness must be allowed in the reference implementation ...


garloff:
Closing this -- having missed a git pull, the commit history looks messy, which is wasting precious reviewer time.
So I'm redoing this one.

PR #318, opened 2023-01-20
Default to Ubuntu 22.04 on management host.

garloff:
It needed a workaround setting the env NEEDRESTART_SUSPEND=1, as apt install -y does not seem sufficient to prevent interaction desires from the system.


garloff:
Thanks, Sven!

PR #320, opened 2023-01-23
New openstack CLI tooling changed network output.

garloff:
Rather than networkname=IP, IP we now have 'networkname': ['IP', 'IP'] as output from server list. Parse both.


garloff:
Probably this was caused by upgrading the mgmtcluster to Ubuntu 22.04, with newer openstack client tools.
Would have hit us sooner or later anyway.
TODO: Double-check that there are no other places where we parse network output in similar ways.

jschoone:
Looks like this came with 5.5.0, see https://docs.openstack.org/releasenotes/python-openstackclient/unreleased.html#relnotes-5-5-0-bug-fixes:

> The networks and properties fields of the server list command will now be rendered as objects. In addition, the power_state field will now be humanized and rendered as a string value when using the table formatter.

jschoone (approved):

PR #322, opened 2023-01-27
Correct occm and ccsi image versions

chess-knight:
Occm image version in the manifest differed from the specified version.

E.g occm version is 1.25.3 but image in the manifest is 'only' 1.25.0. See https://github.com/kubernetes/cloud-provider-openstack/blob/v1.25.3/manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
This also applies to ccsi.

Signed-off-by: Roman Hros <roman.hros@dnation.cloud>

### DIFF
--- a/terraform/files/bin/apply_openstack_integration.sh
+++ b/terraform/files/bin/apply_openstack_integration.sh
@@ -30,6 +30,7 @@ if test -n "$OCCM_VERSION"; then
     if test ! -s $NAME; then
       curl -L https://github.com/kubernetes/cloud-provider-openstack/raw/$OCCM_VERSION/manifests/controller-manager/$name -o $NAME
       echo -e "\n---" >> $NAME
+      # correct occm image version - workaround for the https://github.com/kubernetes/cloud-provider-openstack/issues/2094
       sed -i "s|\(docker.io/k8scloudprovider/openstack-cloud-controller-manager:\).*|\1$OCCM_VERSION|g" $NAME
     fi
   done
--- a/terraform/files/bin/apply_openstack_integration.sh
+++ b/terraform/files/bin/apply_openstack_integration.sh
@@ -30,6 +30,7 @@ if test -n "$OCCM_VERSION"; then
     if test ! -s $NAME; then
       curl -L https://github.com/kubernetes/cloud-provider-openstack/raw/$OCCM_VERSION/manifests/controller-manager/$name -o $NAME
       echo -e "\n---" >> $NAME
+      sed -i "s|\(docker.io/k8scloudprovider/openstack-cloud-controller-manager:\).*|\1$OCCM_VERSION|g" $NAME
     fi
   done
   OCCM=openstack-cloud-controller-manager-ds-$OCCM_VERSION.yaml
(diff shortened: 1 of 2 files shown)


terraform/files/bin/apply_openstack_integration.sh
```diff
@@ -30,6 +30,7 @@ if test -n "$OCCM_VERSION"; then
     if test ! -s $NAME; then
       curl -L https://github.com/kubernetes/cloud-provider-openstack/raw/$OCCM_VERSION/manifests/controller-manager/$name -o $NAME
       echo -e "\n---" >> $NAME
+      sed -i "s|\(docker.io/k8scloudprovider/openstack-cloud-controller-manager:\).*|\1$OCCM_VERSION|g" $NAME
```
garloff, line 33: Well, this pulls the correct image.
But we are papering over an upstream bug, don't we?
We download the deployment manifest from https://github.com/kubernetes/cloud-provider-openstack/raw/v1.25.3/manifests/controller-manager/ yet it references the container image v1.25.0.
So maybe the image has not changed at all? Or the manifest is buggy.
So shouldn't this be filed as a bug upstream? Would you do that?
That said, the image v1.25.3 seems to exist, so the patch does not hurt ...

garloff: Did you care to report this upstream, @chess-knight ?
Or do you disagree with the assessment?

chess-knight: > So maybe the image has not changed at all?

It definitely changed, because I experienced it when I tested octavia proxy protocol option [here](https://github.com/SovereignCloudStack/issues/issues/250#issuecomment-1408187653). So I created this PR and used this branch for testing that.

chess-knight: > So shouldn't this be filed as a bug upstream?

I can create the upstream bug issue, but can they fix the current "bad" manifests - by replacing the image version in the existing occm release/tag? or will we need to merge this anyway?

garloff: I expect we'll need your workaround at least for 1.25.x with 0 < x <= 3, and it probably won't hurt for any x.
Just want to make sure we also have the issue fixed where it should be fixed.

chess-knight: Ok, I created an upstream [issue](https://github.com/kubernetes/cloud-provider-openstack/issues/2094).

garloff: Thanks, Roman!
I suggest we give it a day or two to see if this is acknowledged as a bug upstream or if there is some unexpected wisdom that we fail to grasp.
I would merge this one then, unless the upstream discussion suggests that this would be a bad idea.
/remind-me 2022-02-16

chess-knight: So they did three new releases [v1.24.6](https://github.com/kubernetes/cloud-provider-openstack/releases/tag/v1.24.6), [v1.25.4](https://github.com/kubernetes/cloud-provider-openstack/releases/tag/v1.25.4) and [v1.26.1](https://github.com/kubernetes/cloud-provider-openstack/releases/tag/v1.26.1). with the correct tags in the manifests.

We should probably update [openstack-kube-versions.inc](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/bin/openstack-kube-versions.inc)

garloff: Yes, we should. But let's first merge this one!
Care to open a request for updating to latest k8s-v1.26.1 and latest OCCM bundle?

garloff (approved):
Looks good to me.

terraform/files/bin/apply_openstack_integration.sh
chess-knight: Of course, I will do that.

chess-knight: Here it is #327

PR #323, opened 2023-01-31
Modify LB IP with dns suffix when the proxy protocol is used

chess-knight:
Related to SovereignCloudStack/issues#250 and #184

See also https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#use-proxy-protocol-to-preserve-client-ip

Signed-off-by: Roman Hros <roman.hros@dnation.cloud>

### DIFF
--- a/terraform/files/template/cloud.conf.tmpl
+++ b/terraform/files/template/cloud.conf.tmpl
@@ -7,3 +7,4 @@ application-credential-secret="${appcredsecret}"
 [LoadBalancer]
 manage-security-groups=true
 use-octavia=true
+enable-ingress-hostname=true


terraform/files/template/cloud.conf.tmpl
```diff
@@ -7,3 +7,4 @@ application-credential-secret="${appcredsecret}"
 [LoadBalancer]
 manage-security-groups=true
 use-octavia=true
+enable-ingress-hostname=true
```
garloff, line 10: You have probably seen the logic in apply_nginx_ingress.sh to use kustomize to patch in the openstack annotations to enable the health monitor and the proxy protocol if NGINX_INGRESS_PROXY is set.
I would prefer to implement the hostname setting also via a conditional kustomization, so we only enable it when we need it. Looks like the annotation approach requires an explicit hostname though (and I don't know what it should be set to ...).
Are you sure this setting is harmless when the proxy-protocol is not enabled?

garloff: Does anyone know the answer to the "Are you sure this is harmless ... ?" question?

garloff: Hmm, reading up https://github.com/kubernetes/ingress-nginx/issues/3996, it would seem that an annotation loadbalancer.openstack.org/hostname: nip.io or the default setting you suggest would be needed until (KEP-1860)[https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1860-kube-proxy-IP-node-binding] gets implemented.
Any information on what `enable-ingress-hostname=true` would do when the proxy protocol is not enabled?

chess-knight: `enable-ingress-hostname=true` is used only when the annotation `loadbalancer.openstack.org/proxy-protocol: "true"`. It is written [here](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#load-balancer) and I also found it in the [code](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/openstack/loadbalancer.go#L1822).

garloff: OK, looks like a safe choice then, thanks!

garloff (approved):
/lgtm

Nils98Ar:
It seems that the enhancement KEP-1860 will be included in K8s 1.29: https://github.com/kubernetes/enhancements/pull/4114.

chess-knight:
Yeah, it will be in the alpha stage. But still, OCCM will need to adapt to that

Issue #321, opened 2023-01-25
K8s 1.23.14 (default) deployment fails


chess-knight:
Similar problem to #303.  
Deployment gets stuck during bootstrap of first control plane because coredns container image cannot be pulled.

**Example log of containerd on control plane:**
```
Jan 25 10:46:36 capi-testcluster-control-plane-genc01-dcck4 containerd[545]: time="2023-01-25T10:46:36.830350539Z" level=info msg="PullImage \"registry.k8s.io/coredns:v1.8.6\""
Jan 25 10:46:37 capi-testcluster-control-plane-genc01-dcck4 containerd[545]: time="2023-01-25T10:46:37.554457011Z" level=error msg="PullImage \"registry.k8s.io/coredns:v1.8.6\" failed" error="rpc error: code = NotFound desc = failed to pull and unpack image \"registry.k8s.io/coredns:v1.8.6\": failed to resolve reference \"registry.k8s.io/coredns:v1.8.6\": registry.k8s.io/coredns:v1.8.6: not found"
```

I tested it and it seems that *registry.k8s.io* is used only from [v1.23.15](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#changelog-since-v12314).  
This also applies to [v1.22.17](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#changelog-since-v12216) and [v1.24.9](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#changelog-since-v1248).  
PR #313 only checks major and minor versions, which seems insufficient.

I found that it was is fixed in cluster-api [v1.2.9](https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.2.9) so maybe we do not need this registry patching logic at all. Maybe we also don't need [`imageRepository`](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/template/cluster-template.yaml#L59) in the KubeadmControlPlane spec. This [issue](https://github.com/kubernetes-sigs/cluster-api/issues/7833) in cluster-api is probably also related.

garloff:
Thanks, @chess-knight !
The registry switch upstream is a PITA.
Trouble with capi >= 1.2.8 and omitting the imageRegistry setting is that certain patch levels will be blocked.
[...]

PR #324, opened 2023-02-01
Address #321: Finegrained repo location patching.

garloff:
So we copy the capi knowledge of whether kube-proxy is at the old location (k8s.gcr.io) or at the new one (registry.k8s.io).
As the logic became somewhat involed, split it out into a separate script that can be tested more easily.

### DIFF
--- a/terraform/files/bin/create_cluster.sh
+++ b/terraform/files/bin/create_cluster.sh
@@ -77,10 +77,7 @@ if test "$CONTROL_PLANE_MACHINE_COUNT" -gt 0 &&  grep '^ *OPENSTACK_ANTI_AFFINIT
 fi
 
 # Patch registry location for k8s >= 1.21
-K8S_MAJMIN=$(grep '^KUBERNETES_VERSION:' $CCCFG | sed 's/^KUBERNETES_VERSION: v\([0-9]*\)\.\([0-9]*\).*$/\1\2/')
-if test "$K8S_MAJMIN" -ge 121 && grep 'k8s\.gcr\.io' ${CLUSTERAPI_TEMPLATE} >/dev/null 2>&1; then
-	sed -i 's/k8s\.gcr\.io/registry.k8s.io/g' ${CLUSTERAPI_TEMPLATE}
-fi
+fixup_k8sregistry.sh "$CCCFG" "${CLUSTERAPI_TEMPLATE}"
 
 cp -p "$CCCFG" $HOME/.cluster-api/clusterctl.yaml
 KCCCFG="--config $CCCFG"
--- a/terraform/files/bin/create_cluster.sh
+++ b/terraform/files/bin/create_cluster.sh
@@ -76,7 +76,7 @@ if test "$CONTROL_PLANE_MACHINE_COUNT" -gt 0 &&  grep '^ *OPENSTACK_ANTI_AFFINIT
 	fi
 fi
 
-# Patch registry location for k8s >= 1.21
+# Patch registry location for k8s (~newer than Nov 2022)
 fixup_k8sregistry.sh "$CCCFG" "${CLUSTERAPI_TEMPLATE}"
 
 cp -p "$CCCFG" $HOME/.cluster-api/clusterctl.yaml
--- a/terraform/files/bin/fixup_k8sregistry.sh
+++ b/terraform/files/bin/fixup_k8sregistry.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # fixup_k8s_version.sh
 # Patch $2 (cluster-template.yaml) with fixed up k8s imageRepo if needed
-# This is to find old kube-proxy images on k8s.gcr.io and new ones on registry.k8s.io
+# This is to find old kube-proxy, coredns, ... images on k8s.gcr.io and new ones on registry.k8s.io
 # See https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/util/kubeadm/kubeadm.go#L40
 # (c) Kurt Garloff, Roman Hros, 02/2023
 # SPDX-License-Identifier: Apache-2.0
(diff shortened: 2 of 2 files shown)


garloff:
This addresses #321 .

garloff:
Your reviews would be appreciated!

garloff:
@jcshoone, @batistein, @joshmue, @chess-knight : Care to review this?

joshmue:
I guess this code could break in a few ways e. g. regarding version strings in the kind-of-longterm.

Anyhow, it only needs to work until no supported version is served by the old registry anymore.

So, when setting up a reminder to delete this code as soon it is not needed anymore when 1.25 is EOL (or when 1.24 is EOL?), it LGTM.

terraform/files/bin/fixup_k8sregistry.sh
```diff
@@ -0,0 +1,20 @@
...
+# Patch $2 (cluster-template.yaml) with fixed up k8s imageRepo if needed
+# This is to find old kube-proxy images on k8s.gcr.io and new ones on registry.k8s.io
+# See https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/util/kubeadm/kubeadm.go#L40
+# (c) Kurt Garloff, Roman Hros, 02/2023
+# SPDX-License-Identifier: Apache-2.0
+
+if test -z "$2"; then echo "ERROR: Need clusterctl.yaml cluster-template args" 1>&2; exit 1; fi
+K8SVER=$(grep '^KUBERNETES_VERSION:' "$1" | sed 's/^KUBERNETES_VERSION: v\([0-9.]*\)/\1/')
+K8SMINOR=${K8SVER#*.}
+K8SVER=${K8SVER%%.*}${K8SMINOR%%.*}$(printf %02i 0${K8SMINOR:3:99})
```
chess-knight, line 12: I found that `$(printf %02i 0${K8SMINOR:3:99})` tries to convert number to octal.
When I plugged in v1.24.8 I got `./fixup_k8sregistry.sh: line 12: printf: 08: invalid octal number` or for v1.24.9 I got `printf: 09: invalid octal number`.
For e.g. v1.24.10 I have **12408** in *$K8SVER* so it doesn't work

garloff:
> I guess this code could break in a few ways e. g. regarding version strings in the kind-of-longterm.
> 
> Anyhow, it only needs to work until no supported version is served by the old registry anymore.

I improved it with code that parses v2.x.y (and also v2.x) successfully.
We won't need it for the registry patching, but who knows what else we'll need it for and code tends to get copied. So better make it robust ...

terraform/files/bin/fixup_k8sregistry.sh
garloff: Ack, this was broken, as printf interpreted numbers with leading zeros as octal numbers.
Thanks for finding this one. Fixed by [0272e8a](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/324/commits/0272e8a8a524606186d2b06fc9293bbe3a37299b).

terraform/files/bin/create_cluster.sh
```diff
@@ -77,10 +77,7 @@ if test "$CONTROL_PLANE_MACHINE_COUNT" -gt 0 &&  grep '^ *OPENSTACK_ANTI_AFFINIT
 fi
 
 # Patch registry location for k8s >= 1.21
```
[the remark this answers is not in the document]
garloff: I added a statement that it's newer k8s versions (since Nov 2022).

terraform/files/bin/fixup_k8sregistry.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# fixup_k8s_version.sh
+# Patch $2 (cluster-template.yaml) with fixed up k8s imageRepo if needed
+# This is to find old kube-proxy images on k8s.gcr.io and new ones on registry.k8s.io
```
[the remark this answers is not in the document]
garloff: I improved the comment.

terraform/files/bin/fixup_k8sregistry.sh
garloff: With e4f984bb60d6f99b313c428c9a77b84cf498a6eb actually.

garloff:
This is ready for re-review now.

terraform/files/bin/fixup_k8sregistry.sh
```diff
@@ -0,0 +1,23 @@
...
+# (c) Kurt Garloff, Roman Hros, 02/2023
+# SPDX-License-Identifier: Apache-2.0
+
+if test -z "$2"; then echo "ERROR: Need clusterctl.yaml cluster-template args" 1>&2; exit 1; fi
+K8SVER=$(grep '^KUBERNETES_VERSION:' "$1" | sed 's/^KUBERNETES_VERSION: v\([0-9.]*\)/\1/')
+K8SMINOR=${K8SVER#*.}
+K8SPATCH=${K8SMINOR#*.}
+if test "$K8SPATCH" = "$K8SMINOR"; then K8SPATCH=0; fi
+K8SMINOR=${K8SMINOR%%.*}
+K8SVER=${K8SVER%%.*}$(printf %02i ${K8SMINOR})$(printf %02i ${K8SPATCH})
```
joshmue, line 15: As I found it hard to understand how the version is being read, what line 13 is doing and why, etc.:

Here is another option using yq and kind-of typed bash variables as much as possible (yq is available in this context, right?):

```
readarray -t -d "." K8SVER < <(yq eval '.KUBERNETES_VERSION | sub("v", "")' "$1" | tr -d '\n')
K8SMAJOR=${K8SVER[0]}
K8SMINOR=${K8SVER[1]}
K8SPATCH=${K8SVER[2]}
K8SVER=${K8SMAJOR}$(printf %02i ${K8SMINOR})$(printf %02i ${K8SPATCH})
```

Regarding actual comparison of versions: Of course, 1) reading the semver parts, 2) reformatting them to a special string and 3) using the string as integer to compare it against other special strings/integers, is pretty prone to errors.
So of course, using a semver library/binary would be optimal.
But all edge cases I can come up with, which could cause problems with the existing script are *probably* unlikely to happen in the near future: E. g. Kubernetes changing its release cadence to have fewer minor releases with many more patch releases, leading to versions like `1.30.101` which would cause problems here.

Anyways: Seems to work for this use case for now; If it should be really robust, IMHO considering to use a semver library like in the cluster-api code is the way to go. Alternatively, I imagine, such comparison could also be done without a library, when being done in Python: Just creating a simple semver class with attributes major/minor/patch and a simple `__lt__(self, other)` implementation attached. I'd be surprised if the same is possible with bash ;)

garloff: > As I found it hard to understand how the version is being read, what line 13 is doing and why, etc.:

The bash shell substitution is not easy to read:
`${VAR%x}` returns `$VAR` with the shortest matching `x` removed from the end and the original `$VAR` if no `x` matches.
`${VAR#x}` does the same thing at the beginning, `%%` and `##` remove the longest matches at the end and beginning respectively.

If we are afraid of cumbersome code, I would turn to python instead rather than using another tool.
Yes, yq is available, but I'm not sure that everyone is familiar with it.

I would tend to keep it with hard-to-read bash for now, I'm not afraid of 3-digit patchlevels ...

garloff: Line 13 specifically handles the case that there is no match, so we succeed parsing v1.24 as v1.24.0 => 12400. We would otherwise misparse to 12424.

PR #326, opened 2023-02-21
Fix needrestart configuration

matofeder:
`Needrestart` is part of the `apt-get upgrade` process in Ubuntu 22.04. By default, it is set to an "interactive" mode which causes the interruption of scripts. The interactive mode is applied when the new kernel version is available after the upgrade process or when upgraded services need to restart.

### Issue 

 Needrestart causes that after the new kernel is released and/or a new version of some upgraded service is released the needrestart daemon opens an interactive dialog  and interrupts scripts e.g. as follows: 

- A new kernel version is available
  Screenshot from 2023-02-20 14-34-19

- Services should be restarted
  image

### Steps to reproduce the issue 

Create the management host (`make create`) using TF variable: `image = "Ubuntu 22.04"` .

### PR
  
This PR adds a custom configuration file that overrides mentioned settings and ensures the following:
- Kernel hints are printed only to STDERR 
- Services are restarted automatically if needed

Notes:

- It is necessary to configure the needrestart daemon within the `cloud-config` script as the `cloud-config` script  
upgrades packages `package_upgrade: true`
- `NEEDRESTART_SUSPEND=1` variable has been removed as we have global settings in place now
- IMO automatic restart of services after an upgrade is a better approach than ignoring the needed restart of services by  suspending the  needrestart daemon


chess-knight (approved):

garloff (approved):
This looks like the cleaner approach, indeed!
And yes, restarting services during the bootup process (when the package_upgrade is running) is the better choice; just rebooting would not always be trivial to get right, but we don't do it.
THANKS!

PR #327, opened 2023-02-23
Update to latest k8s and occm patch

chess-knight:
This PR increases k8s versions to latest patch as well as occm patch versions as discussed in #322.

### DIFF
--- a/terraform/files/bin/openstack-kube-versions.inc
+++ b/terraform/files/bin/openstack-kube-versions.inc
@@ -3,12 +3,12 @@
 # (c) Kurt Garloff <kurt@garloff.de>, 3/2022
 # SPDX-License-Identifier: Apache-2.0
 # Images from https://minio.services.osism.tech/openstack-k8s-capi-images
-k8s_versions=("v1.18.20" "v1.19.16" "v1.20.12" "v1.21.14" "v1.22.17" "v1.23.15" "v1.24.9" "v1.25.5" "v1.26.0")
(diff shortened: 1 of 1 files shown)


terraform/files/bin/openstack-kube-versions.inc
```diff
@@ -3,12 +3,12 @@
 # (c) Kurt Garloff <kurt@garloff.de>, 3/2022
 # SPDX-License-Identifier: Apache-2.0
 # Images from https://minio.services.osism.tech/openstack-k8s-capi-images
-k8s_versions=("v1.18.20" "v1.19.16" "v1.20.12" "v1.21.14" "v1.22.17" "v1.23.15" "v1.24.9" "v1.25.5" "v1.26.0")
+k8s_versions=("v1.18.20" "v1.19.16" "v1.20.15" "v1.21.14" "v1.22.17" "v1.23.16" "v1.24.10" "v1.25.6" "v1.26.1")
```
garloff, line 6: I did not know that 1.20.x had proceeded that far.
Trouble is that osism does not build and publish node images for these old k8s versions any more.
Maybe we should drop at least 1.18 and 1.19 from this list ...
Anyhow, we should do this with another PR, not here, so this is good!

garloff (approved):
Thanks, Roman!

terraform/files/bin/openstack-kube-versions.inc
chess-knight: Thank you.
I know that osism builds only 1.21+  now, but I wanted to have this list "complete".

PR #329, opened 2023-02-24
Fix kustomization of nginx ingress

### LINKED ISSUE #328: Kustomize errors when building nginx manifests
I always face the error
```
Error: invalid Kustomization: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field Kustomization.patches of type types.Patch
```
This comes from the `kustomize build` lines from `apply_nginx_ingress.sh`
Using `kubectl kustomize` seems to fix this issue, I will create a PR.

jschoone:
Fixes #328

### DIFF
--- a/terraform/files/bin/apply_nginx_ingress.sh
+++ b/terraform/files/bin/apply_nginx_ingress.sh
@@ -21,9 +21,9 @@ if test ! -s base/nginx-ingress-controller-${NGINX_VERSION}.yaml; then
 fi
 ln -sf nginx-ingress-controller-${NGINX_VERSION}.yaml base/nginx-ingress-controller.yaml
 if test "$NGINX_INGRESS_PROXY" = "false"; then
-	kustomize build nginx-monitor > ~/$CLUSTER_NAME/deployed-manifests.d/nginx-ingress.yaml || exit 3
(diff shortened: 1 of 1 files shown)


chess-knight (approved):
I tested it and it seems it works.

batistein (approved):

garloff:
As per discussion in the Container Team meeting:
* kustomize has changed upstream and we probably need to adjust our kustomization for it to pass
* kubectl kustomize still has an older version of kustomize included which is the reason for it to still work
* As soon as the kubectl included kustomize gets revved, we'd break again

Suggested way forward:
(a) Adjust the kustomization files such that they are processed successfuly through kustomize v5
Only if this fails:
(b) Find a way to reliably get and pin an older version of kustomize (via kubectl or directly)

garloff:
Reading through https://github.com/kubernetes-sigs/kustomize/pull/4911, I guess that we just need to replace patches: with patchesStrategicMerge: in the kustomization files to be compatible with kustomize v5 (and v3).

garloff:
Better to fully update the syntax and use `resources` instead of `base` and `patches:\n  - path: FILENAME` instead of `patchesStrategicMerge:\n  - FILENAME`.

PR #340, opened 2023-02-28
Cert-Manager is v1.11.0 by default now

### LINKED ISSUE #334: Update cert-manager to latest ahead of R4.
As user that opts-in to get cert-manager with the cluster, I'd like to get the latest stable version deployed.

chess-knight:
Up from v1.9.1. Fixes #334


garloff (approved):
/lgtm, thanks!

PR #341, opened 2023-02-28
Upgrade to nginx_ingress controller v1.6.4

### LINKED ISSUE #333: Update nginx ingress to latest ahead of R4.
As k8s user, I would like to get the latest version of the nginx ingress controller provisioned with the k8s cluster (if I don't opt out).

chess-knight:
Up from v1.3.0. Fixes #333


garloff (commented):
Looks good.
Have you tested that the the LB created this way works?

chess-knight:
> Looks good. Have you tested that the the LB created this way works?

I tested it with the following [proxy procedure](https://github.com/SovereignCloudStack/issues/issues/250#issuecomment-1409916779).

garloff (approved):
Looks good, thanks!

PR #347, opened 2023-03-01
Fix deletion of management server floating IP in cleanup script

### LINKED ISSUE #342: Failed to delete floating_ip in cleanup script
Cleanup script `./cleanup/cleanup.sh` logs the following (s/XXX.XXX.XXX.XXX/floating IP 4 address/) when trying to delete the floating IP associated with the management node:
```
openstack floating ip delete 'XXX.XXX.XXX.XXX']}
Failed to delete floating_ip with name or ID ''XXX.XXX.XXX.XXX']}': No FloatingIP found for '213.131.230.96']}
```

Floating IP is wrongly filtered in the  [resourcelist](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/cleanup/cleanup.sh#L106) function. This function returns brackets `]}` on the end of the floating IP which causes an issue.

matofeder:
The cleanup script's function `resourcelist` wrongly filtered the floating IP of the management server. The function returned brackets `]}` on the end of the floating IP string which causes an issue. 

This PR extends `resourcelist` function arguments with an argument that allows a custom definition of `sed` expression which filters the output of the function. The custom `sed` expression is then used to filter the management server floating IP without unwanted brackets on the end.

Closes #342


batistein (approved):

garloff (commented):
The way networks are listed has changed between old openstackclient-5.3 and -5.5 (if my memory serves) and I guess this causes the need for this change.
If we can't easily support both, we might want to document the needed version.

jschoone:
> The way networks are listed has changed between old openstackclient-5.3 and -5.5 (if my memory serves) and I guess this causes the need for this change. If we can't easily support both, we might want to document the needed version.

In addition: see #320

matofeder:
> The way networks are listed has changed between old openstackclient-5.3 and -5.5 (if my memory serves) and I guess this causes the need for this change. If we can't easily support both, we might want to document the needed version.

@garloff Thank you for pointing it out (your memory serves well) . `sed` expression has been adjusted to support both openstack-client response formats, see 5bbb206cf71c3e29eab505c3f94f2894c2893d98

PR #348, opened 2023-03-01
Adjust to new kustomize syntax.

garloff:
Use resources rather than base.
Insert path: before filenames referenced in patches. Reference: https://github.com/kubernetes-sigs/kustomize/pull/4911 Supercedes: https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/329


garloff:
Your reviews would be welcome!

jschoone (approved):
Just tested and works for me.
I closed #329 without merging.

Nils98Ar:
@garloff Can this also be merged into maintained/v4.x? Otherwise we would need to switch to main.

garloff:
Thanks, @Nils98Ar, this bugfix should definitely go to v4.x and also v4.0.x.
I have already merged (cherry-picked) this kustomize fix into both branches.

Honestly, I was a bit unsure how much those maintained branches are in use and we have not given them as much attention lately.
I opened issue #369 and #370 to discuss and track what we backport in terms of bugfixes and (in case of v4.x) minor updates.

PR #349, opened 2023-03-01
Fix cluster API load balancer lookup in the cleanup script

### LINKED ISSUE #344: Cleanup script does not consider cluster name when look for cluster API load balancer
When e.g. two CAPI clusters (with different names) want to share one OpenStack project and then one of them should be removed, the cleanup script wrongly removes all cluster API load balancers that contain `clusterapi` string. The cleanup script does not consider the cluster name, see [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/cleanup/cleanup.sh#L122).

Suggested solution:
Consider cluster name in the cluster API load balancer filtering

matofeder:
The cleanup script did not consider a cluster name when it looked for load balancers created by CAPI. 
This might cause the cleanup script wrongly removes all CAPI load balancers from the OpenStack project,
when e.g. two CAPI clusters (with different names) want to share one OpenStack project.

This PR adds a cluster name to the filter of CAPI load balancers.

Closes #344

### DIFF
--- a/terraform/cleanup/cleanup.sh
+++ b/terraform/cleanup/cleanup.sh
@@ -111,15 +111,15 @@ fi
 echo "Deleting cluster $CLUSTER"
 # cleanup loadbalancers
 if test -n "$NOCASCADE"; then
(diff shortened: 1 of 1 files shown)


garloff (approved):
Thanks, Matej, Roman!
This looks good to me.

PR #350, opened 2023-03-01
Format tables in README.md

matofeder:
This PR does the formatting of all tables in the README.md file. This PR does not change the content of tables, just prettifies their markdown format.

This PR should simplify further editing of tables in the README.md file. 

TIP: Creating tables with hyphens and pipes can be tedious. This could be automatized by e.g. IDE extensions or by tools like http://markdowntable.com/

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,63 +434,62 @@ been added by the SCS project..
 
 Parameters controlling the Cluster-API management node (capi mgmt node) creation:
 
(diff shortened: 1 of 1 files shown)


garloff (approved):
Thanks Matej.
This definitely makes reading the MD files easier.
I was indeed too lazy to do this before and just relied on users looking at the (HTML) rendered versions to parse the tables.

PR #352, opened 2023-03-03
Upgrade calico to v3.25.0

### LINKED ISSUE #305: calico_version ineffective
As noted in the commit log to PR #304 (commit ce61c247445d1adfcc9ced3226012c7037746f7b), we currently have a parameter calico_version in the terraform which does control the version of the installed calicoctl binary, but *not* the more important kubernetes manifest calico.yaml. This is highly confusing and was not intended.

To fix this, do either:
- Drop calico_version (and just pick one that is validated -- might have a weak dependence on the k8s version)
- Make it effective by downloading the wanted version of calico.yaml (like we do for many other services). In this case, we should also add it to clusterctl.yaml (like the other services).

### LINKED ISSUE #331: Update calico to latest stable ahead of R4 (and test!)
As k8s user, I want to benefit from the latest stable calico features.

chess-knight:
Up from v3.24.1. Fixes #331

Calico version is effective now. It downloads correct version of manifests. Fixes #305


garloff (approved):
Thanks!
/lgtm

PR #354, opened 2023-03-03
Add prefix to security groups allow-ssh and allow-icmp

### LINKED ISSUE #343: Non-prefixed security groups `allow-ssh` and `allow-icmp` may cause unwanted behavior when two CAPI clusters (with different names) want to share one OpenStack project
Security groups `allow-ssh` and `allow-icmp` are created by the terraform script [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/neutron.tf#L3) and then referenced in cluster template [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/template/cluster-template.yaml#L153) and [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/template/cluster-template.yaml#L199).

In a special case when e.g. two management nodes and then two CAPI clusters (with different names) want to share one OpenStack project the `allow-ssh` and `allow-icmp` are created two times. These sec. groups have the same names but have different IDs. 
As the CAPI provider looks for sec. group by its name, not by ID, the cluster node could have the following "doubled" sec. groups association:

```
| security_groups             | name='allow-icmp' 
|                             | name='allow-icmp'
|                             | name='allow-ssh'   
|                             | name='allow-ssh'   
```
 This is an unwanted behavior that may break e.g. cleanup process, where the terraform wants to remove sec. group used by another cluster.    

Steps to reproduce this issue:

1. Create two management nodes and two k8s clusters with a custom `prefix`  and `testcluster_name`.
Bundles have to share one OpenStack project, i.e. cloud_provider "test1" and "test2" should point to the same OS project. 

test1 env. file
```
cat >terraform/environments/environment-test1.tfvars <<EOF
prefix            = "test1"
cloud_provider    = "test1"

availability_zone = "nova"
external          = "ext01"
kind_flavor       = "SCS-2V:4:20"
controller_flavor = "SCS-2V:4:20"
worker_flavor     = "SCS-2V:4:20"

testcluster_name  = "test1"
EOF
```

test2 env. file
```
cat >terraform/environments/environment-test2.tfvars <<EOF
prefix            = "test2"
cloud_provider    = "test2"

availability_zone = "nova"
external          = "ext01"
kind_flavor       = "SCS-2V:4:20"
controller_flavor = "SCS-2V:4:20"
worker_flavor     = "SCS-2V:4:20"

testcluster_name  = "test2"
EOF
```

make create
```
OS_CLOUD=test1 make create
OS_CLOUD=test2 make create
```

2. Check k8s cluster node sec. groups:
```
$ openstack server show  <worker-node-name> -f json | jq '.security_groups'
[
  {
    "name": "allow-icmp"
  },
  {
    "name": "allow-icmp"
  },
  {
    "name": "allow-ssh"
  },
  {
    "name": "allow-ssh"
  },
  {
    "name": "k8s-cluster-default-test1-secgroup-worker"
  }
]
```

Suggested solution:
Mentioned sec. groups cloud be prefixed with the "prefix" variable

matofeder:
Security groups allow-ssh and allow-icmp are created by the terraform script and then referenced in the cluster template. In a special case when e.g. two management nodes and then two CAPI clusters (with different names) want to share one OpenStack project, the allow-ssh, and allow-icmp are created two times. These sec. groups have the same names but have different IDs. As the CAPI provider looks for sec. group by its name, not by ID, the cluster node could have the "doubled" sec. groups association. This is an unwanted behavior that may break e.g. cleanup process, where the terraform wants to remove sec. group used by another cluster.

This PR adds a "prefix" to the allow-ssh and allow-icmp sec. groups that distinguish their usage in a single OpenStack project.

Closes #343

### DIFF
--- a/terraform/cleanup/cleanup.sh
+++ b/terraform/cleanup/cleanup.sh
@@ -174,7 +174,7 @@ if test "$FULL" == "1"; then
 	#cleanup router ${CAPIPRE}-
 	cleanup_list router "" "" "$RTR"
 	cleanup "security group" ${CAPIPRE}-mgmt
-	cleanup "security group" allow-
+	cleanup "security group" ${CAPIPRE}-allow-
 	cleanup keypair ${CAPIPRE}-keypair
 	cleanup "application credential" ${CAPIPRE}-appcred
 	#cleanup volume pvc-
--- a/terraform/files/template/cluster-template.yaml
+++ b/terraform/files/template/cluster-template.yaml
@@ -150,8 +150,8 @@ spec:
         name: ${CLUSTER_NAME}-cloud-config
         kind: Secret
       securityGroups:
-        - name: allow-ssh
-        - name: allow-icmp
+        - name: ${PREFIX}-allow-ssh
+        - name: ${PREFIX}-allow-icmp
         - name: ${PREFIX}-${CLUSTER_NAME}-cilium
 ---
 apiVersion: cluster.x-k8s.io/v1beta1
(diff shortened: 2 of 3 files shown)


garloff (approved):
Right, we have prefix to allow for multiple management hosts in one project and the security groups could cause trouble (in cleanup, as you say).
Thanks for the improvement!

PR #355, opened 2023-03-05
Feat/protect etcd backup

garloff:
Fix timer activation and chmod 0600 for the backup to better protect against container escapes.


matofeder:
This is a rather general comment regarding ETCD:

Defragmentation of a live ETCD member blocks the system from reading and writing data while rebuilding its states, see the [docs](https://etcd.io/docs/v3.5/op-guide/maintenance/#defragmentation). Therefore, in the case of single control plane node (i.e. single instance of ETCD), the defragmentation could cause a temporary outage of ETCD, hence k8s API may fail to read/write to it. 

In case of ETCD cluster (multiple instances, e.g. 3) it is recommended to defragment  ETCD members one at a time. This could be achieved e.g. by `--cluster` or `--endpoints` option.  The current timer `OnCalendar=*-*-* 02:30:00` may or may not fulfill that.  

In addition, it is a good practice to defragment the current ETCD leader last, to prevent the triggering of multiple leader elections.

I would say that this step-by-step defragmentation could be performed from the management node, but for sure this should be addressed in a separate issue.  Of course, this could be addressed only if we want to achieve ETCD defragmentation without any ETCD cluster outage.

terraform/files/template/cluster-template.yaml
```diff
@@ -114,7 +118,7 @@ spec:
           RandomizedDelaySec=15m
 
           [Install]
-          WantedBy=multi-user.target
+          WantedBy=timers.target
```
matofeder, line 121: The changes work as expected, but the `etcd-defrag.service` is not started (only enabled), hence the defragmentation does not work until the control plane node is rebooted. 

Steps to reproduce:
1. Create a management node and cluster
```
OS_CLOUD=test1 make create
```
2. SSH to the control plane node
3. Check if the timer is scheduled
```
# systemctl list-timers --all
NEXT                        LEFT         LAST                        PASSED              UNIT                         ACTIVATES                     
Wed 2023-03-08 11:40:05 UTC 2h 8min left Wed 2023-02-01 11:38:10 UTC 1 months 4 days ago motd-news.timer              motd-news.service             
...         
n/a                         n/a          n/a                         n/a                 etcd-defrag.timer            etcd-defrag.service 
```

I do not know if the above is intended. If not this could be fixed e.g. by `systemctl enable --now` (in postKubeadmCommands)

matofeder (approved):

PR #356, opened 2023-03-05
Change default k8s version to v1.25.x (x=latest).

garloff:


chess-knight (approved):
I tested it and all pods are running without restarts. I also passed sonobuoy quick mode.

Issue #246, opened 2022-08-19
Add ability to limit access to k8s API


garloff:
Currently, the load balancer in front of the kube-apiserver (:6443) allows connections from everywhere, so you can manage the created workload k8s cluster from anywhere (if you have a kubeconfig with the right certificate).
This has some downsides:
* Cluster owners may be nervous exposing the kubeAPI to the world, "only" protected by a client certificate. In particular, the hostile world may attempt to connect, causing lots of connections and a certain load with the SSL/TLS connection setups.
* Considering to significantly prolong the timeouts until the load balancer cuts idle connections (to avoid broken *internal* connections, e.g. from kubelet to kubeAPI) may not be such a good idea with lots of potentially evil external connection attempts, eating resources at the load balancer.

The idea is that we introduce a new parameter, allowing the use to specify an array of IP ranges (CIDRs). Default would be 0/0, as today.
If a range different from 0/0 is specified, the implementation would need to add the NODE_CIDR range to it (so cluster-internal connections to the API work) as well as the FIPs of the management cluster[*], so the kubectl connections from it work.

[*] Should we allow the IPs used for SNAT of outgoing connections from this cloud provider? This would ensure access even in case of changes in the management cluster ... On the other hand, any evil party with an account on this cloud could again talk to the API server again and "only" be protected by not having the needed client cert. If a user really cares to not want 0/0 access, I would assume that we should be restrictive here, so my weak opinion on this is "no".

garloff:
Maybe a value of "null" should be supported, allowing ONLY the automatically added ranges (NODE_CIDR and mgmt node FIP).

garloff:
This would be implemented in the loadbalancer listener settings in front of kubeapi:
`openstack loadbalancer listener set --allowed-cidr 10.0.0.0/8 --allowed-cidr 172.16.0.0/12 --allowed-cidr 192.168.0.0/16 --allowed-cidr MGMTHOSTI/32`
Of course we could also allow our own range of outgoing (NATted) IP addresses.
Plus anything the customer wants.

garloff:
Is relevant and should be addressed.

garloff:
See https://cluster-api-openstack.sigs.k8s.io/clusteropenstack/configuration.html#restrict-access-to-the-api-server

PR #360, opened 2023-03-05
Feat/add kubapi restrictions

garloff:
This implements #246.


garloff:
Sidenote: The patches here are relative to a branch that I constructed to work against the latest, not-yet-merged state.
This makes it easier to review.
Technically, a merge to the branch does not require a review but would be useless.
So, please look at the implementation here.
When we are all fine, we would probably have to rebase against the then current state of main and require a formal review in order to really merge.

chess-knight (commented):
I found that this is not supported by OVN provider. See https://cluster-api-openstack.sigs.k8s.io/clusteropenstack/configuration.html#restrict-access-to-the-api-server or https://docs.openstack.org/octavia/xena/user/feature-classification/index.html#operation_allowed_cidr. I suggest to mention this somewhere.

terraform/files/bin/apply_kubeapi_cidrs.sh
```diff
@@ -0,0 +1,60 @@
...
+{
+	if test -z "$1" -o "$1" = "null"; then return 0; fi
+	if test "$1" = "[]" -o "$1" = "[ ]" -o "$1" = "[  ]"; then return 0; fi
+	return 1
+}
+
+get_own_fip()
+{
+	NETS=$(openstack server list --name "${PREFIX}-mgmtcluster" -f value -c Networks)
+	FIP=${NETS##*, }
```
chess-knight, line 19: I found the same problem as [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/issues/342) with `]}` at the end

garloff: Good point, need to make this work with openstackclient versions >= and < 5.5.

We don't offer ovn LB in front of kube-api yet, the cloud.conf settings only affect occm created LBs (k8s objects of service type LoadBalancer).
So we can mention it, but it's not something that anyone would run into without extra manual tuning.
Still: good catch!

chess-knight: > We don't offer ovn LB in front of kube-api yet, the cloud.conf settings only affect occm created LBs (k8s objects of service type LoadBalancer).

Thank you for pointing this out. CAPO has to be configured specially to use ovn lb-provider.

garloff: Comment has been added and (more importantly) the regexp adjusted to work with both pre-5.5 and 5.5+ openstackclien network output format.t

garloff:
I will close this in favor of #368 which is a rebase against current main.

PR #362, opened 2023-03-05
Use new v2 flavor names.

### LINKED ISSUE #361: Default to new v2 flavor names
As KaaS operator with an SCS-compatible IaaS, I want to have defaults that work.
This means that we want [v2 flavor names](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v2-flavor-naming.md) as defaults in the capi environment.

garloff:
Fixes #361.


jschoone (approved):
LGTM

PR #364, opened 2023-03-07
Pin flux version

### LINKED ISSUE #335: Update flux to the latest version ahead of R4
As k8s user that opts-in to get flux2 deployed along with the cluster, I'd like to get the current stable version.

chess-knight:
Fixes #335


terraform/files/bin/create_cluster.sh
```diff
@@ -173,8 +173,12 @@ fi
 
 # Flux2
 DEPLOY_FLUX=$(yq eval '.DEPLOY_FLUX' $CCCFG)
-if test "$DEPLOY_FLUX" = "true"; then
-  KUBECONFIG=${KUBECONFIG_WORKLOADCLUSTER} flux install || exit $?
+if test "$DEPLOY_FLUX" = "true" -o "${DEPLOY_FLUX:0:1}" = "v"; then
+  FLUX_INSTALL_OPTS="--timeout 10m0s"
+  if test "${DEPLOY_FLUX:0:1}" = "v"; then
```
garloff, line 178: Shouldn't we pin flux to some known-to-be-good version also for `true`?
Like we do in the `bootstrap.sh` script? Or does the flux binary default to the same version that it has? But then, multiple clusters may want to use multiple flux versions.
Imagine a testcluster deployment via terraform, pinning some brand-new version, which results in a very new binary. Should a secondary cluster with `DEPLOY_FLUX=true` not install the known-to-be-good default version (0.40.2)?

garloff (changes requested):

terraform/files/bin/create_cluster.sh
chess-knight: Flux binary installs the same version of flux.
You can't install a different flux version as binary is(it can only differ in patch version).

garloff: OK.

garloff (approved):

PR #366, opened 2023-03-08
Pin cilium

### LINKED ISSUE #332: Update cilium to latest version ahead of R4 (and test!)
As k8s user, I want to have the latest cilium features available, especially also the ability to have the ability to beta-test the gateway API that is included.

chess-knight:
Fixes #332


garloff:
This one is a bit more complex...
Thanks! 
Looks good at first sight. You seem to have found all places where changes are needed and rightly have moved the location of USE_CILIUM. Maybe we need to export the variable? I'll double-check tomorrow in front of a computer and then do a formal review.

chess-knight:
Thanks for the first sight review.

> Maybe we need to export the variable?

I think it is sourced by `. ~/.capi-settings` in both scripts.

garloff:
I looked at the implementation in a bit more detail.
One disadvantage of the way we have done the pinning now is that USE_CILIUM has now become a global option (per capi management server) as opposed to a per-cluster option that it was before. From that perspective it even is a regression.

Maybe we need to split setting of choosing the version of the cilium binary, which gets installed on the management server and is thus global (per capi mgmt server) from the decision whether or not to use cilium (and not calico) and if so which cilium version, which should continue to be per-cluster.

How strong is the dependence of cilium binary with deployed cilium artifacts (containers, ...)?
Maybe the cleanest way would be to deploy versioned cilium binaries on the capi mgmt server, so we call the appropriate one per cluster.

chess-knight:
> I looked at the implementation in a bit more detail. One disadvantage of the way we have done the pinning now is that USE_CILIUM has become a global option (per capi management server) as opposed to a per-cluster choice cilium vs. calico that it was before. From that perspective it even is a regression.

I understand but, isn't `CALICO_VERSION` also a global option?

> Maybe we need to split setting of choosing the version of the cilium binary, which gets installed on the management server and is thus global (per capi mgmt server) from the decision whether or not to use cilium (and not calico) and if so which cilium version, which should continue to be per-cluster.
> How strong is the dependence of cilium binary with deployed cilium artifacts (containers, ...)? Maybe the cleanest way would be to deploy versioned cilium binaries on the capi mgmt server, so we call the appropriate one per cluster.

It makes sense. Maybe we can omit the cilium binary version setting completely and install some fixed version or the latest stable because I read that it should be backwards compatible [here](https://cilium.io/blog/2021/05/20/cilium-110/) and also it should support a [newer version](https://github.com/cilium/cilium-cli#releases) of the cilium, but can we rely on it? The same way is done hubble binary now, we install the latest stable version and it is also [backwards compatible](https://github.com/cilium/hubble#releases).

terraform/variables.tf
```diff
@@ -172,7 +172,7 @@ variable "etcd_unsafe_fs" {
 variable "git_branch" {
   description = "k8s-cluster-api-provider git branch to be checked out on mgmtserver"
   type        = string
-  default     = "master"
+  default     = "main"
```
jschoone, line 175: fyi: Looks like this branch is based on an older version since it was already changed. But I can't find the origin.

jschoone:
For me this feels way to complex. Is that format commonly used?

garloff:
Superceded by merged #385.

PR #371, opened 2023-03-10
Add --grouping=false to clusterctl describe cluster.

garloff:
The output is much nicer, at least for clusters that are of moderate size.


jschoone (approved):
Lgtm

PR #385, opened 2023-03-14
Pin cilium and binaries

### LINKED ISSUE #332: Update cilium to latest version ahead of R4 (and test!)

chess-knight:
Fixes #332

It is a different approach than #366. *USE_CILIUM* is used for the cilium version per cluster. There is additional variable *cilium_binaries* which represents the cilium and hubble CLI versions for the management host.


garloff (approved):
Thanks Roman, this looks really good to me.
* We can select the binaries, but have good defaults
* We can chose the cilium version, and again have a good default
* We can opt-in to cilium per cluster

Issue #384, opened 2023-03-14
etcd defrag + backup: Avoid too many leader changes


garloff:
As k8s cluster user, I want the k8s control plane to always be responsive, stable and safe.

We have a nightly job to defragment etcd and back it up on all control plane nodes, randomized a bit, so the defragmentation does not happen all at the same time.
This has been in existence for many months already, but due to a missing `--now` in `systemctl enable`, it has not really been active before.
As @matofeder [points out](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/355#issuecomment-1459872544), the defragmentation may block access to etcd for a while (seconds on typically sized etcd DBs), causing etcd leader changes (on multi-node etcd clusters) or temporary kube-api failures (on single-node etcd clusters).

Things to consider:
1. Possibly stronger protection from concurrent defragmentation on multiple etcd nodes by scheduling all etcd nodes from the leader instead of relying on the configured randomness in the timer start time.
2. Scheduling the leader etcd defragmentation last, as it will likely cause a leader change and we want to minimize these. (Starting with the leader would cause several leader changes ...)
3. Skipping the leader's defragmentation (for up to a week or infinitely?) to cause less leader changes?
4. Skipping defragmentation on single-node etcd installations?
5. Leaving this disabled for R4 and do more real-world tests before R5. (This is not without risk either, we have seen heavily fragmented etcds causing trouble in real-life already.)

garloff:
To point 1: It should be noted that the backup and fstrim operations can not be done centrally but would need a local job on the non-leader control-plane nodes, possibly triggered via ssh.

garloff:
Point 2 is easy to do.

garloff:
Points 3+4: If we decide to simply skip defrag on the etcd leader (infinitely), we'd cover both of these points with a single change.

garloff:
Your feedback is welcome, especially @matofeder, @chess-knight, @batistein, @janiskemper, @ajfriesen, @flyersa, @curx, @fynluk, @mxmxchere.

batistein:
@guettli this is what we talked about. maybe we could contribute here

guettli:
I think the best solution would be an official solution from etcd.io. 

Maybe it is enough to update the docs: I created an issue for that: https://github.com/etcd-io/etcd/issues/15477

[...]

PR #387, opened 2023-03-14
Quickfix for #384: Don't defrag etcd on leader.

garloff:
This is the quick-fix: We just don't do defragmentation on the etcd leader. This avoids etcd service interruption on single-node etcd clusters and spurious leader changes on multi-node ones.

Note that this is the intermediate step until we have a more complete solution as depicted in
https://github.com/SovereignCloudStack/k8s-cluster-api-provider/issues/384#issuecomment-1468290695

### DIFF
--- a/terraform/files/template/cluster-template.yaml
+++ b/terraform/files/template/cluster-template.yaml
@@ -78,6 +78,25 @@ spec:
         kubeletExtraArgs:
           cloud-provider: external
     files:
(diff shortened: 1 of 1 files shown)


matofeder (approved):
LGTM!

PR tested and works as expected:
- defragmentation, backup, fstrim are executed only on non-leader members

PR #391, opened 2023-03-16
Move calico version to the clusterctl.yaml

chess-knight:
It was mentioned in #305 but not done in #352.


garloff (approved):
Looks good, thanks.

PR #392, opened 2023-03-16
Comment on AZs, treat YQ3 as old yq.

garloff:
Minor usability improvements ...


terraform/Makefile
```diff
@@ -27,12 +27,12 @@ endif
...
 YQVERSION=$(shell yq --version | $(SED) -e 's/^yq //' -e 's/^.*version //')
 YQMAJOR=$(shell YQV="$(YQVERSION)"; echo "$${YQV%%.*}")
-ifeq ($(YQMAJOR),2)
-  YQ=yq --yaml-output
-  YQIN=
-else
+ifeq ($(YQMAJOR),4)
   YQ=yq eval
   YQIN=-
+else
```
chess-knight, line 33: I tried it with yq in version 2.4.1, and also 3.4.1 but I am still getting `unknown flag: --yaml-output`

chess-knight (commented):
I tested it with latest(v4) and also v3 and v2 but without success.

terraform/neutron.tf
```diff
@@ -73,6 +73,7 @@ resource "openstack_networking_router_v2" "router_mgmt" {
   name                    = "${var.prefix}-rtr"
   description             = "router for mgmtcluster (managed by terraform)"
   external_network_id     = data.openstack_networking_network_v2.external.id
+  # comment this out if your cloud does not have network AZs
```
chess-knight, line 76: *terraform fmt* doesn't like it
```diff
$ terraform fmt -check -diff terraform/neutron.tf 
terraform/neutron.tf
--- old/terraform/neutron.tf
+++ new/terraform/neutron.tf
@@ -70,9 +70,9 @@
 }
 
 resource "openstack_networking_router_v2" "router_mgmt" {
-  name                    = "${var.prefix}-rtr"
-  description             = "router for mgmtcluster (managed by terraform)"
-  external_network_id     = data.openstack_networking_network_v2.external.id
+  name                = "${var.prefix}-rtr"
+  description         = "router for mgmtcluster (managed by terraform)"
+  external_network_id = data.openstack_networking_network_v2.external.id
   # comment this out if your cloud does not have network AZs
   availability_zone_hints = [var.availability_zone]
 }
```

terraform/Makefile
garloff: Looks like we need to differentiate differently between the two yq variants out there.

terraform/neutron.tf
chess-knight: Maybe we can place it next to availability_zone.

terraform/Makefile
```diff
@@ -27,12 +27,12 @@ endif
 
 YQVERSION=$(shell yq --version | $(SED) -e 's/^yq //' -e 's/^.*version //')
 YQMAJOR=$(shell YQV="$(YQVERSION)"; echo "$${YQV%%.*}")
-ifeq ($(YQMAJOR),2)
-  YQ=yq --yaml-output
-  YQIN=
-else
+ifeq ($(YQMAJOR),4)
```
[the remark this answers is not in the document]
chess-knight:
```suggestion
ifeq ($(YQMAJOR),v4)
```

terraform/Makefile
chess-knight: There are also two potential sources of `yq`:

[mikefarah/yq](https://github.com/mikefarah/yq) - can be downloaded as binary or installed via snap
- I don't see `--yaml-output` flag
- I see `eval` command in the *v4*

[kislyuk/yq](https://github.com/kislyuk/yq) - installed with pip and it has **jq** dependency
- I see `--yaml-output` flag
- I don't see `eval` command

terraform/neutron.tf
jschoone: > Maybe we can place it next to availability_zone.

Looks like this is the only solution to keep the indentation.
But the line is quite long then.
I'm not sure what is "better practice", spontaneous I would prefer the "wrong" indentation by `terraform fmt` over a long line.

jschoone: @garloff @chess-knight after some more research I saw Gruntwork recommends [120 column limit](https://docs.gruntwork.io/guides/style/terraform-style-guide/#120-column-limit) in which we still are, if we place the comment next to the attribute.
That would change my mind and I would go with it. Some opinions?

chess-knight: I am OK with that, also when I look on the terraform [style conventions](https://developer.hashicorp.com/terraform/language/syntax/style) I can see e.g. `count = 2 # meta-argument first` line there.

jschoone: Done

terraform/Makefile
chess-knight: I tested the current version and I think that it is sufficient.
```bash
ifeq ($(YQMAJOR),v4)
  YQ=yq eval
  YQIN=-
else
  YQ=yq --yaml-output
  YQIN=
endif
```
When the [mikefarah/yq](https://github.com/mikefarah/yq) is used, we fall into the **ifeq v4** case, because the output of *--version* looks like this `yq (https://github.com/mikefarah/yq/) version v4.33.3`
When the [kislyuk/yq](https://github.com/kislyuk/yq) is used, we fall into the **else** case, because the output of *--version* looks like this `yq 3.2.2`

If we want to support [mikefarah/yq](https://github.com/mikefarah/yq) **< v4**, we need to introduce more complexity to this and it also means that we will support deprecated and unmaintained tools, because [v2](https://mikefarah.gitbook.io/yq/v/v2.x/) and also [v3](https://mikefarah.gitbook.io/yq/v/v3.x/) is deprecated and unsupported.

What do you think @jschoone @garloff?

jschoone: Hi @chess-knight, thanks for picking this up again.
This looks good. I tested a bit with switching between [mikefarah/yq](https://github.com/mikefarah/yq) v4.33 and [kislyuk/yq](https://github.com/kislyuk/yq) 3.2.2 and it all worked.
I also agree to not support the old versions of [mikefarah/yq](https://github.com/mikefarah/yq)

PR #399, opened 2023-03-17
Fix k9s installation

chess-knight:
x86_64 was replaced by amd64 in [v0.27.0](https://github.com/derailed/k9s/releases/tag/v0.27.0)


garloff (approved):
Thanks for spotting this.
/lgtm

PR #401, opened 2023-03-18
Update etcd defrag and backup

### LINKED ISSUE #384: etcd defrag + backup: Avoid too many leader changes

matofeder:
This PR updates etcd defragmentation and backup script as follows:
- Script exits without any defragmentation/backup/trim action if:
  - It is executed on non leader etcd member
  - It is executed on etcd cluster with some unhealthy member
  - It is executed on single member etcd cluster

- Conditions above could be skipped and the script execution could be forced by the optional arguments:
  - `--force-single`
  - `--force-unhealthy`
  - `--force-nonleader`
 
- Script then defragments the etcd cluster as follows:
  - Defrag the non leader etcd members first
  - Change the leadership to the randomly selected and defragmentation completed etcd member
  - Defrag the local (ex-leader) etcd member

- Script then backup & trim local (ex-leader) etcd member

Note:
This script executes etcdctl commands like `etcdctl move-leader` or `etcdctl endpoint status --cluster` which were introduced in etcdctl version 3.3.0. The previous etcdctl client was installed as an `apt` package. The latest etcdctl version available in Ubuntu 20.04 repositories is v3.2.26, hence this PR also installs the desired version of etcdctl client. Etcdctl client is then used for etcd DB maintenance tasks.

Closes #384

### DIFF
--- a/doc/Maintenance_and_Troubleshooting.md
+++ b/doc/Maintenance_and_Troubleshooting.md
@@ -113,3 +113,64 @@ or have a look at the nodes:
 If you fall into some Kubernetes specific issues after a successful cluster
 creation, go through the official [Kubernetes](https://kubernetes.io/docs/tasks/debug/debug-cluster/)
 troubleshooting guide.
+
+## Etcd maintenance
+
+[Etcd](https://etcd.io/) is a highly-available key value store used as Kubernetes'
+backing store for all cluster data. This section contains etcd related maintenance
+notes from SCS k8s-cluster-api-provider project perspective.
+
+For further information about etcd maintenance visit an official [etcd maintenance guide](https://etcd.io/docs/v3.5/op-guide/maintenance/)
+and/or [Kubernetes etcd operating guide](https://kubernetes.io/docs/tasks/administer-cluster/configure-upgrade-etcd/).
+
+### Defragmentation and backup
+
+Etcd storage can become fragmented over time, which causes the performance to decrease
+and may even cause premature out-of-space conditions. For this we had included a
+maintenance script that regularly defragments and backups the database. The script, called
+`etcd-defrag.sh` is located in the `/root` directory of each control plane node.
+It is executed through the systemd service unit file `etcd-defrag.service` and scheduled
+to run each day at 02:30:00 using the `etcd-defrag.timer` systemd timer.
+
+The defragmentation strategy is inspired by the [etcd-defrag-cronjob](https://github.com/ugur99/etcd-defrag-cronjob/) and
+[practices recommended](https://docs.openshift.com/container-platform/4.9/scalability_and_performance/recommended-host-practices.html#automatic-defrag-etcd-data_recommended-host-practices) by the OpenShift project.
+Note that the proposed strategy could be changed in a future version based on results from
+related [upstream issue #15477](https://github.com/etcd-io/etcd/issues/15477) which wants to define
+an official solution on how to defragment etcd cluster.
+
+The `etcd-defrag.sh` script checks multiple conditions before the actual defragmentation as
+follows:
+- The script should not be executed on non leader etcd member
+- The script should not be executed on etcd cluster with some unhealthy member
+- The script should not be executed on single member etcd cluster
+
+These pre-flight checks should ensure, that the defragmentation does not cause temporary
+etcd cluster failures and/or unwanted etcd leader changes. They also prevent executing
+the script on a single control-plane node cluster. Single-node etcd clusters are not
+made for long-term operation. As a workaround, however, you can scale up to three
+control-plane nodes overnight from time to time.
+
+After all pre-flight checks passed the etcd cluster defragmentation is performed as follows:
+- Defragment the non leader etcd members first
+- Change the leadership to the randomly selected and defragmentation completed etcd member
+- Defragment the local (ex-leader) etcd member
+
+At the end of the defragmentation script, the local (ex-leader) etcd member is backed up
+and trimmed. Backup is saved and then compressed in the control plane `/root` directory.
+You can find it here: `/root/etcd-backup.xz`. File system trim is performed by the `fstrim`
+command that discards unused blocks on a filesystem which could increase write performance
+on the long run and also release unused storage.
+
+All mentioned pre-flight checks could be skipped by the optional arguments that force
+defragmentation despite potential failures. Optional arguments are:
+- `--force-single` (allows to execute defragmentation on single member etcd cluster)
+- `--force-unhealthy` (allows to execute defragmentation on unhealthy etcd member)
+- `--force-nonleader` (allows to execute defragmentation on non leader etcd member)
+
+**We do not recommend executing the manual defragmentation** using the optional arguments above.
+If you are aware of potential issues, you can access the control plane node and
+execute the defragmentation script manually as follows:
+
+```bash
+/root/etcd-defrag.sh [--force-single] [--force-unhealthy] [--force-nonleader]
+```
--- a/doc/Maintenance_and_Troubleshooting.md
+++ b/doc/Maintenance_and_Troubleshooting.md
@@ -158,7 +158,10 @@ At the end of the defragmentation script, the local (ex-leader) etcd member is b
 and trimmed. Backup is saved and then compressed in the control plane `/root` directory.
 You can find it here: `/root/etcd-backup.xz`. File system trim is performed by the `fstrim`
 command that discards unused blocks on a filesystem which could increase write performance
-on the long run and also release unused storage.
+on the long run and also release unused storage. Cluster admins are not supposed to log
+in to the cluster nodes (neither control plane nor workers) and thus won't access or use
+these backup files. The local backups on these nodes however can prove useful however
+in a disaster recovery scenario.
 
 All mentioned pre-flight checks could be skipped by the optional arguments that force
 defragmentation despite potential failures. Optional arguments are:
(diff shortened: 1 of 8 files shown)


terraform/files/template/cluster-template.yaml
```diff
@@ -83,17 +83,65 @@ spec:
...
+            exit 0
+          fi
+
+          # Get all etcd members with their endpoints, IDs, and leader status
+          declare -a MEMBERS
+          while read MEMBER; do
+            MEMBERS+=( "$MEMBER" )
+          done < <($ETCDCTL endpoint status --cluster)
+
+          if test $${#MEMBERS[@]} = 1; then
```
matofeder, line 113: Note: $${#MEMBERS[@]} - double `$` here and below as a escape character for `clusterctl generate` command

garloff:
We'll be doing this improvement past R4 (but likely pull it into the maintained branches).

chess-knight:
I fixed the script in f5b897529f784543a4a808c88733a26a4a58aeb9. There was `{#` - jinja template comment, which cloud-init doesn't like. I also fixed the installation of *etcdctl* tool.

I have done multiple tests with this defrag script. For 1 control-plane and also for 3 control-plane cluster everything works as expected.

garloff:
We do dev work on the `main` branch normally and can then (selectively) backport to maintained branches.

doc/Maintenance_and_Troubleshooting.md
```diff
@@ -113,3 +113,64 @@ or have a look at the nodes:
...
+backing store for all cluster data. This section contains etcd related maintenance
+notes from SCS k8s-cluster-api-provider project perspective.
+
+For further information about etcd maintenance visit an official [etcd maintenance guide](https://etcd.io/docs/v3.5/op-guide/maintenance/)
+and/or [Kubernetes etcd operating guide](https://kubernetes.io/docs/tasks/administer-cluster/configure-upgrade-etcd/).
+
+### Defragmentation and backup
+
+Etcd storage can become fragmented over time, which causes the performance to decrease
+and may even cause premature out-of-space conditions. For this we had included a
```
joshmue, line 129: > which causes the performance to decrease
> and may even cause premature out-of-space conditions

I think these claims should be backed by referring to etcd documentation or some case study.

[etcd defragmentation docs](https://etcd.io/docs/v3.5/op-guide/maintenance/#defragmentation) AFAIU only say that while defragmentation is useful for giving back disk space to the host filesystem, the space is still available to etcd - without notions about performance or out-of-space conditions.

Let's back this up with some references or remove it.

(Of course, doing defragmentation before doing a file-based backup would have advantages like smaller backup sizes)

garloff:
> We do dev work on the `main` branch normally and can then (selectively) backport to maintained branches.

I have created `maintained/v5.0.x` and `maintained/v5.x` branches now.
But of course, we merge to main first and then, possibly even after some delay for real-world testing, merge the "backports" into maintained branches as we judge appropriate. (It's good for a PR against main to already mention that this might be material for backports -- cf. Cc: stable@kernel.org.)

doc/Maintenance_and_Troubleshooting.md
joshmue: > If the defragmentation is not performed in a long run the out-of-space condition may occur on the host filesystem which may lead to performance degradation on both host and then potentially on etcd too. I did not find any shiny case study that shows the correlation between the host's out-of-space and the host's performance degradation as this could depend on many factors like the type of disk etc.

Well, if the disk really runs out of space, this may cause indeed `unpredictable etcd cluster behavior`, not just performance problems.

> It is true that defragmentation is useful for giving back disk space to the host filesystem. If the defragmentation is not performed in a long run the out-of-space condition may occur on the host filesystem

That implies that etcd does not reuse space for itself, causing a steady increase of disk space usage. That would be a contradiction to the content of the etcd documentation.

Under https://etcd.io/docs/v3.5/op-guide/maintenance/#space-quota, the docs say that defragmentation is required when the cluster ran out of space earlier - to bring the cluster in an operable state again.
But even if doing periodic defragmentation, there is AFAIU nothing preventing the keyspace to explode *between the defragmentation cycles*.

According to my understanding, reasonable sizings/rate-limits/compactions/quotas are required to prevent out-of-space conditions as well as possible; with defragmentation playing mostly a minor role when recovering/backing-up clusters.

matofeder: I got your point that defragmentation does not prevent out-of-space conditions, but mainly compactions should be performed before actual defragmentation. Also, it could be unclear that etcd reuses space for itself. Therefore, I removed the statement from docs 41d003bba43d0362a93e369f911f7421769b3a8c.

joshmue: Thanks!

> I got your point that defragmentation (alone) does not prevent out-of-space conditions, but compaction + defragmentation do this.

(Even though I'd still argue that AFAIU, defragmentation does not really prevent out-of-space conditions *at all*, unless other things on the same disk use space that should have been planned to be available to etcd)

garloff: @joshmue - I agree that we should also start to look at rate limits ... if we want to substantially increase the operational safety.

garloff (commented):
I will push the simplifications (removal of etcdctl choice) and the comments on cluster-admins NOT supposed to log into the k8s nodes into the branch.

doc/Maintenance_and_Troubleshooting.md
```diff
@@ -113,3 +113,63 @@ or have a look at the nodes:
...
+
+After all pre-flight checks passed the etcd cluster defragmentation is performed as follows:
+- Defragment the non leader etcd members first
+- Change the leadership to the randomly selected and defragmentation completed etcd member
+- Defragment the local (ex-leader) etcd member
+
+At the end of the defragmentation script, the local (ex-leader) etcd member is backed up
+and trimmed. Backup is saved and then compressed in the control plane `/root` directory.
+You can find it here: `/root/etcd-backup.xz`. File system trim is performed by the `fstrim`
+command that discards unused blocks on a filesystem which could increase write performance
```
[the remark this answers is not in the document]
matofeder: addressed by fixup 0526d5d68776e9351d18b123af0739b1c5dd700f

terraform/environments/environment-default.tfvars
```diff
@@ -15,6 +15,7 @@ cilium_binaries      = "<v0.aa.bb;v0.xx.yy>"  # defaults to "v0.13.1;v0.11.2"
 kubernetes_version   = "<v1.XX.XX>"		  # defaults to "v1.25.x"
 kube_image_raw       = "<boolean>"      # defaults to "true"
 calico_version       = "<v3.xx.y>"	# defaults to "v3.25.0"
+etcdctl_version      = "<v3.x.yy>"      # defaults to "v3.5.7"
```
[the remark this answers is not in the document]
matofeder: addressed by fixup 1b6920adbb801403e70d9ae5906894064a0f613d

doc/configuration.md
```diff
@@ -60,6 +60,7 @@ Parameters controlling the cluster creation:
...
 | `testcluster_name`               | (cmd line)                                | SCS        | `testcluster`                            | Allows setting the default cluster name, created at bootstrap (if `controller_count` is larger than 0)                                                                                                     |
 | `restrict_kubeapi`               | `RESTRICT_KUBEAPI`                        | SCS        | `[ ]`                                    | Allows restricting access to kubernetes API by list of CIDRs. Empty list (default) means public, `[ "none" ]` means internal access only.                                                                  |
+| `etcdctl_version`                | `ETCDCTL_VERSION`                         | SCS        | `v3.5.7`                                 | Version of the etcdctl client that is used for etcd DB maintenance tasks                                                                                                                                   |
```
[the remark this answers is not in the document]
matofeder: addressed by fixup 1b6920adbb801403e70d9ae5906894064a0f613d

garloff (approved):
Looks good to me now, thanks.
PS: Let's test this in main for a while before backporting it to maintained/v5.x and v5.0.x.

guettli:
FYI: There is a new tool for etc defragmentation: https://github.com/ahrtr/etcd-defrag  It is is written by a core maintainer of etcd.

Issue #325, opened 2023-02-08
Use OVN provider LB when available in front of ingress


garloff:
As an ingress user, I want to see real client IPs in the ingress.
To make this work without proxy-protocol workarounds, we must ensure 
- [ ] that we use externalTrafficPolicy: local (otherwise we see the kube-proxy forwarded source IPs)
- [ ] have a health-monitor enabled (otherwise the nodes without the nginx will receive and drop traffic)
- [ ] have a loadbalancer that does preserve the client IPs (by working on layer 3 or similar ways to steer the flows)

The octavia load balancer implemented by the OVN provider does preserve the client IPs nicely.

To use it, we can set lb-provider=ovn in [loadbalancer] sertion of cloud.conf, of course only after testing that it is supported by the underlying infra.

See also issues/#268 for figuring out how to make health monitors work.

garloff:
OVN LB is not ready yet:
* It only support TCP Health-Mon while OCCM unconditionally wants an HTTP health-mon
* The health-mon is ineffective when the service is access via FIP (it work for VIPs)
See [issues/#268](https://github.com/SovereignCloudStack/issues/issues/268)

garloff:
PR #351 added the code to use the OVN provider LB in front of ingress (or other services).
As it does not work yet, it's currently guarded by an `--allow-preview-features` switch.

garloff:
Sidenote: Testing zed release (on a cloud from OSISM):
* Manually creating OVN LB with VIP in the same network as the members, healthmon for backend members, attaching FIP, the access via the FIP still shows the same issue not correctly taking out the broken members of the LB rotation, only the VIP accesses.
* On the other hand, using the openstack-health-monitor with -LP ovn, the accesses via FIP *are* taken out when the http servers on the members are killed. The VIP here lives in a different net than the members (and they are connected via a router).
Whether it's really the network location of the VIP that makes the difference is unknown, there may be other differences.

garloff:
The health-mon not working for the floating IPs *should* be fixed in zed since early March, see 
commit https://review.opendev.org/c/openstack/ovn-octavia-provider/+/873860
(commit https://review.opendev.org/c/openstack/ovn-octavia-provider/+/875595 against Zed)
which fixes https://bugs.launchpad.net/neutron/+bug/1997418/
which appears to be a duplicate of https://bugs.launchpad.net/neutron/+bug/1956035.

[...]

PR #407, opened 2023-04-11
Use custom container for OCCM when using OVN LB.

garloff:
We currently need a patched OCCM that falls back to TCP HealthMon when HTTP is not supported (OVN LB).
Detect this and create own version of k8s occm manifest for this.

With this, when calling `create_cluster.sh --allow-preview-features` with setting `USE_OVN_LB_PROVIDER: true` or `auto`, we will have a very nicely working LB in front of a service, fully supporting `externalTrafficPolicy: local` by handling both the failover as well as the client IP visibility -- without resorting to proxy-protocol trickery. This requires SCS R4 (OSISM v5.0.0 or better v5.1.0) IaaS (OpenStack zed with current OvS/OVN).

This is what we have been working on for a while, see [issues/#298](https://github.com/SovereignCloudStack/issues/issues/298), and #325.

Of course this can only be an intermediary state. A patch needs to go upstream into OCCM ... before this can be considered without `--allow-preview-features`. But do the integration now, so the usefulness can be demonstrated and the feature be tested without manual lumbing.


terraform/files/bin/apply_openstack_integration.sh
```diff
@@ -24,14 +24,26 @@ else
 fi
 echo "# Install external OpenStack cloud provider $OCCM_VERSION to $CLUSTER_NAME"
 
+OCCM_RAW=$OCCM_VERSION
+OCCM_OVN=v1.26.202
+if grep '^lb\-provider=ovn' ~/$CLUSTER_NAME/cloud.conf >/dev/null 2>&1; then
+  OCCM_VERSION=$OCCM_OVN
+  OCCM_RAW=v1.26.2
```
chess-knight, line 31: What about version compatibility, will it also work for Kubernetes < 1.26? E.g. we use CCMR_VERSION(roles, role-bindings) below, but it can be in a different minor version.

garloff: Hi Roman,
we are prudent to use the recommended OCCM versions that match the k8s version. That said, I have not yet been able to observe incompatibility issues with newer OCCMs since the 1.20 times or so ...
I have tested the 1.26 OCCM successfully with 1.25 clusters.
The feature is experimental anyhow for now (until we can get a patch merged upstream), maybe we should just add a note to the documentation that the customer OCCM build has only been tested with k8s-1.25 and -1.26.

Care to do some ingress testing on gx-scs with this?

chess-knight: Thank you for that note in the documentation, I will do ingress testing.

chess-knight: Ok, I did some testing with a simple app with ingress in front of it, and it doesn't work as expected for me. It seems like the health monitor doesn't work because sometimes I get `Connection timed out` when the request is going from outside of the cluster. From within the cluster, it works.

My setup is the default test cluster, with `USE_OVN_LB_PROVIDER: true` and `--allow-preview-features`.
I am not able to investigate if it is really a health monitor issue or not. But it seems like 2/3 of requests are not balanced correctly. But in the Horizon everything looks ok, only one node is online.

chess-knight: From the upstream, I can see that the fix is probably included in the 3.1.0 version of openstack/ovn-octavia-provider project but I don't know if it is included in the latest stable zed, because [here](https://docs.openstack.org/releasenotes/ovn-octavia-provider/zed.html) is only 3.0.0 mentioned. But [here](https://releases.openstack.org/zed/#zed-ovn-octavia-provider) the 3.1.0 is mentioned as the most recent version and also [here](https://docs.openstack.org/ovn-octavia-provider/zed/) is 3.1.x.

garloff: Christian @berendt, what is the best way to find out what version of octavia-ovn-provider is in OSISM-5.1.0?

garloff: @chess-knight @jschoone - The question is whether we should merge this change despite upstream OVN provider not being fully ready - preparing for the time when it becomes ready and allowing to test and expose the issue.
Of course, we also need to then push for the patch (or a variant of it that is configurable?) to go into upstream OCCM.

garloff: > Christian @berendt, what is the best way to find out what version of octavia-ovn-provider is in OSISM-5.1.0?

The answer is researched in https://github.com/osism/issues/issues/499.

garloff:
Upstream issue for the OCCM patch: https://github.com/kubernetes/cloud-provider-openstack/issues/2225

garloff:
This needs retesting.

chess-knight:
Hello @garloff,
I read https://github.com/kubernetes/cloud-provider-openstack/pull/2234 and https://github.com/kubernetes/cloud-provider-openstack/pull/2489 and according to the comments there, it should work out of the box without these changes(because TCP type is the default protocol in the code `Type:       string(port.Protocol),`).

So I gave it a try, and I tried it on the latest main(k8s v1.27.5, occm v1.27.2, gx-scs cloud) and **it worked** for me. Every request is routed correctly. From my point of view, all past OpenStack bugs are already resolved in the cloud and that's why it works.

---

During testing, I found only one small bug in our current code:
when I try to create a new cluster with the *create_cluster.sh --allow-preview-features* we have `handle_ovn_lb.sh "$CLUSTER_NAME" || exit 1` there, but  `handle_ovn_lb.sh` returns 2 so `exit 1` is called, that's because handle_ovn_lb.sh calls use_ovn() and it calls set_cfg_octavia() which returns 2. The second run is fine, because `handle_ovn_lb.sh` returns 0.

garloff:
Nice!
So let's fix the return value thing and get rid of the preview status and adjust the docs accordingly. At least for main and maintained/6.x.

garloff:
A custom container is no longer required.

PR #413, opened 2023-04-28
Make openstack instance create timeout configurable

### LINKED ISSUE #382: Set CLUSTER_API_OPENSTACK_INSTANCE_CREATE_TIMEOUT
Hi :wave: , 
I had some trouble with a k8s update from 1.23.x to 1.24.x, as the capo-controller-manager was running into the "machine create timeout" which is set to 5min by default. I felt like setting `CLUSTER_API_OPENSTACK_INSTANCE_CREATE_TIMEOUT` to 10 does the trick. I am wondering, whether it is possible to set this variable with the tools provided in this repo and would really appreciate any comment :slightly_smiling_face:.

jschoone:
Fixes #382


doc/configuration.md
```diff
@@ -60,6 +60,7 @@ Parameters controlling the cluster creation:
...
 | `testcluster_name`               | (cmd line)                                | SCS        | `testcluster`                            | Allows setting the default cluster name, created at bootstrap (if `controller_count` is larger than 0)                                                                                                     |
 | `restrict_kubeapi`               | `RESTRICT_KUBEAPI`                        | SCS        | `[ ]`                                    | Allows restricting access to kubernetes API by list of CIDRs. Empty list (default) means public, `[ "none" ]` means internal access only.                                                                  |
+| `capo_instance_create_timeout`   | `CLUSTER_API_OPENSTACK_INSTANCE_CREATE_TIMEOUT` | SCS  | `10`                                     | Time to wait for an OpenStack machine to be created (in minutes)                                                                                                                                           |
```
garloff, line 63: Shouldn't the provenance of this parameter be attributed to capo?
We did not invent it, not even the name ...

terraform/variables.tf
```diff
@@ -208,3 +208,10 @@ variable "restrict_kubeapi" {
   type        = list(string)
   default     = []
 }
+
+
+variable "capo_instance_create_timeout" {
+  description = "time to wait for an openstack machine to be created (in minutes)"
+  type        = number
+  default     = 10
```
garloff, line 216: Is it wise to increase the default timeout so much?
Normally instance creation is well below 1min -- only if a port assignment fails and the VM needs to be rescheduled is when we see times in the 5 - 6min range.
I guess I would have chose 6mins as default allowing for 1 resched in an otherwise well-working cloud without making us wait for the error result too long.
(We will need longer timeouts once we spawn bare metal machines, but this is not the case yet, and I would probably then document this.)

garloff (commented):
Looks good to me beyond the two minor comments.

doc/configuration.md
jschoone: Yes

terraform/variables.tf
jschoone: I used the same value the user did who filed the issue, since I didn't see any disadvantages. If there are, we probably should stay at 5 and let the user who needs more adjust it.

I don't think we will spawn bare metal machines using CAPO.

matofeder (approved):
LGTM! 

PR has been tested: 
1. Create a cluster and set custom `capo_instance_create_timeout`
```bash
TF_VAR_capo_instance_create_timeout=8 OS_CLOUD=test make create
```
2. Verify `CLUSTER_API_OPENSTACK_INSTANCE_CREATE_TIMEOUT` env variable
```bash
$ k -n capo-system get deployments.apps/capo-controller-manager -oyaml | yq .spec.template.spec.containers[0].env
- name: CLUSTER_API_OPENSTACK_INSTANCE_CREATE_TIMEOUT
  value: "8"
```

PR #416, opened 2023-05-09
Fix command not found in Makefile

chess-knight:
Fix also sonobuoy.sh usage

### DIFF
--- a/terraform/Makefile
+++ b/terraform/Makefile
@@ -75,7 +75,7 @@ show: init
 
 clean: init
 	@if test -e ./.deploy.MGMTCLUSTER_ADDRESS.$(ENVIRONMENT); then source ./.deploy.MGMTCLUSTER_ADDRESS.$(ENVIRONMENT); \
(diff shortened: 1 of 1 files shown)


garloff (commented):
Hmm, shouldn't we then call '~/bin/script' rather than 'bin/script' ? I meant to have '~/bin' in 'PATH' . Apparently that's not universally the case. You rely on '.' being part of 'PATH', I'm not sure this is always true.

chess-knight:
> Hmm, shouldn't we then call `~/bin/script` rather than `bin/script` ? I meant to have `~/bin` in `PATH` . Apparently that's not universally the case. You rely on `. `being part of `PATH`, I'm not sure this is always true.

I found this PR #150 where the same approach is used. Maybe we can add `~/` to all occurrences in the Makefile to make it more robust. What do you think @garloff?

garloff:
> aybe we can add `~/` to all occurrences in the Makefile to make it more robust. What do you think @garloff?

Yes, let's do this.

chess-knight:
> Yes, let's do this.

See 3c80118fff77cf2dd2ce9e552f01301d19c34925

garloff (approved):
Thanks, looks good to me.

Issue #157, opened 2022-03-08
external net should be autodetected


garloff:
Most clouds have exactly one external network for floating IP addresses.
This could be perfectly autodetected.
So we should allow an empty default for the `external` setting and use it for autodetection.

Issue #423, opened 2023-06-16
Support v3 flavors better


garloff:
As a k8s-capi user (self-managed clusters), I want default flavors that work on all SCS IaaS that comply to SCS-compatiable IaaS v3.

In particular this means:
(1) Leveraging the newly mandatory SCS-2V-4-20s flavors with local SSD for the control plane (etcd)
(2) Allowing the usage of diskless flavors by allocating a volume on the fly as needed for those flavors

Number (2) may be a duplicate of #421.

garloff:
Part (1) is trivial and already implemented. Branch feat/flavors-v3
Part (2) - See https://cluster-api-openstack.sigs.k8s.io/topics/crd-changes/v1alpha4-to-v1alpha5.html#changes-to-rootvolume

chess-knight:
@flyersa, @garloff, @Nils98Ar, this should probably be closed. Also, it has a R5 milestone!

And what about #421?

garloff:
Closed by #424.

PR #424, opened 2023-06-16
Feat/flavors v3

garloff:
Use v3 flavors: SCS-2V-4-20s is default now,
We support diskless flavors now and patch the cluster-template.yaml as needed to create the rootVolume.
We also remove it again as needed.


garloff:
This addressed #423

garloff:
For the terraform check fail, see #425.

garloff:
Would be nice to get a review on this @jschoone @chess-knight @Nils98Ar @flyersa.

terraform/mgmtcluster.tf
```diff
@@ -36,14 +40,27 @@ locals {
...
+data "openstack_images_image_ids_v2" "images" {
+  name = var.image
+  sort = "updated_at:desc"
+}
+
 resource "openstack_compute_instance_v2" "mgmtcluster_server" {
   name              = "${var.prefix}-mgmtcluster"
-  image_name        = var.image
+  # image_name        = var.image
   flavor_name       = var.kind_flavor
```
chess-knight, line 51: I am wondering, what will happen in underlying openstack infra when flavor with the disk is defined here. I tried it with `kind_flavor = "SCS-2V:4:20"` and in the horizon I see that flavor with 20GB disk is used and also 30 GB volume attached on /dev/sda. inside mgmt VM I see:
```
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda1        29G  5.5G   24G  20% /
```

chess-knight (commented):
I tested it with gx-scs environment with following changes:
```terraform
kind_flavor       = "SCS-2V:4:20"
controller_flavor = "SCS-2V:4"
worker_flavor     = "SCS-2V:4"
```
It works for me! The 20GiB volumes are created and attached

garloff:
> I tested it with gx-scs environment with following changes:
> ```terraform
> kind_flavor       = "SCS-2V:4:20"
> controller_flavor = "SCS-2V:4"
> worker_flavor     = "SCS-2V:4"
> ```
> 
> It works for me! The 20GiB volumes are created and attached

And deleted again on VM deletion according to my tests ...

terraform/files/bin/fixup_flavor_volumes.sh
```diff
@@ -0,0 +1,66 @@
...
+	else
+		echo -e "# Volume for worker node disk\nWORKER_ROOT_DISKSIZE: $DISKWORK" >> $1
+	fi
+	cp -p $2 $2.orig
+	kustpatch.sh ~/kubernetes-manifests.d/add-vol-to-worker.yaml <$2.orig >$2
+else
+	sed -i 's/^\(WORKER_ROOT_DISKSIZE\)/#\1/' $1
+	cp -p $2 $2.orig
+	kustpatch.sh ~/kubernetes-manifests.d/rmv-vol-from-worker.yaml <$2.orig >$2
+fi
```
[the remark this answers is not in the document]
garloff: Addressed in [b08b149](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/424/commits/b08b149ebed05486024d52ee760686371d5dc385)

terraform/mgmtcluster.tf
garloff: The disk that comes with the flavor normally is just ignored then? OK, I did not know this ... but it's good, as it makes everything robust here.

chess-knight: It looks like that, but I don't know what happens on the underlying OpenStack layer, specifically what happens with the disk specified in the flavour. If it is ignored, it is good but otherwise can have a problem, don't you think so too?

chess-knight: Ok, I investigated a little bit and found this [flavors docs](https://docs.openstack.org/nova/zed/user/flavors.html#overview).
There is a **Root Disk GB** section which states that `When booting from a persistent volume it is not used.` so most likely we can believe that.

There is also mention of [os_compute_api:servers:create:zero_disk_flavor](https://docs.openstack.org/nova/zed/configuration/policy.html#nova) policy rule, which probably works:
```bash
$ openstack server create --flavor "SCS-2V:4" --image "ubuntu-capi-image-v1.25.11" --network "capi-net" test
Only volume-backed servers are allowed for flavors with zero disk. (HTTP 403)
```
The same message I get if I used it with this repository(main branch), the capo controller reports the same error.

garloff: Great, so we can move forward!

terraform/mgmtcluster.tf
```diff
@@ -10,9 +10,13 @@ resource "openstack_identity_application_credential_v3" "appcred" {
   unrestricted = true
 }
 
+data "openstack_networking_network_v2" "extnet" {
+  external = true
+}
```
chess-knight, line 15: I assume that this fixes #157

garloff: Yes

terraform/files/bin/flavor_disk.sh
```diff
@@ -0,0 +1,56 @@
...
+if test -n "$2"; then
+	IMAGE=`openstack image show $2 -f json`
+	if test $? != 0; then exit -3; fi
+	ISIZE=`echo "$IMAGE" | jq '.min_disk'`
+else
+	ISIZE=20
+fi
+CPU=`echo "$FLAVOR" | jq '.vcpus'` #  | tr -d '"'
+RAM=`echo "$FLAVOR" | jq '.ram'`
+RAM=$(((RAM+64)/1024))
```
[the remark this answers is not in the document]
garloff: It's a way of rounding to avoid that a few missing MBs don't cause the next flavor (with double the size) to be required.

terraform/files/bin/fixup_flavor_volumes.sh
```diff
@@ -0,0 +1,66 @@
...
+CTRLFLAVOR=$(yq eval '.OPENSTACK_CONTROL_PLANE_MACHINE_FLAVOR' $1)
+WORKFLAVOR=$(yq eval '.OPENSTACK_NODE_MACHINE_FLAVOR' $1)
+
+flavor_disk.sh "$CTRLFLAVOR" "$UBU_IMG_NM"
+DISKCTRL=$?
+flavor_disk.sh "$WORKFLAVOR" "$UBU_IMG_NM"
+DISKWORK=$?
+
+if test $DISKCTRL -ge 128; then echo "ERROR $((DISKCTRL-256)) using ctrlplane flavor $CTRLFLAVOR for image $UBU_IMG_NM"; exit 1; fi
+if test $DISKWORK -ge 128; then echo "ERROR $((DISKWORK-256)) using worker flavor $WORKFLAVOR for image $UBU_IMG_NM"; exit 1; fi
```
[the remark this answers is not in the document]
garloff: Two reasons:
(a) I would not want node images that require root volumes > 100GiB
(b) The way the wanted size is returned from the helper uses the exit code where >= 128 indicates an error. The exit code is a single byte ...

We could of course use a different mechanism (e.g. echo the needed size to stdout and read it `$(helper)`), but this way it could be kept a bit simpler. If you see the need for larger volumes, please explain, so we can change this.

chess-knight (approved):
Thank you for the comments.
Lgtm!

Issue #425, opened 2023-06-16
terraform set-output deprecated


garloff:
```Run terraform fmt -check
/home/runner/work/_temp/0c4a498b-331c-4001-a577-b3ed9b8b3f0b/terraform-bin fmt -check
mgmtcluster.tf
variables.tf

Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
Error: Terraform exited with code 3.
Error: Process completed with exit code 1.
```

garloff:
And I can not find saveState or setOutput, so the terraform check error confuses me.

PR #426, opened 2023-06-16
Use tf-1.3.6 for github workflow.

garloff:


garloff:
Fixes #425

.github/workflows/check-terraform-syntax.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Setup terraform
         uses: hashicorp/setup-terraform@v1
         with:
-          terraform_version: 1.0.5  # renovate: datasource=github-releases depName=hashicorp/terraform
+          terraform_version: 1.3.6  # renovate: datasource=github-releases depName=hashicorp/terraform
```
matofeder, line 30: I would like to suggest pinning terraform version to [1.4.6](https://releases.hashicorp.com/terraform/1.4.6/)  instead of 1.3.6. 
Two reasons: 
1. 1.4.6 is the latest patch o 1.4 and 1.5 is quite new (June 12) without any patches so far. 
2. I have a real use case for TF 1.4's  `terraform_data` resource feature:
    - I would like to use   `terraform_data`  resource instead of deprecated `null_resource` in https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/432 

Then, (probably in a separate PR) I would like to suggest increasing the TF version in [main.tf](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/main.tf#L7) as well. 
IMO the GitHub workflow should test the code with the same version of TF as is defined in `main.tf`.

.github/workflows/check-terraform-syntax.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Setup terraform
         uses: hashicorp/setup-terraform@v1
```
chess-knight, line 28:
```suggestion
        uses: hashicorp/setup-terraform@v2
```

.github/workflows/check-terraform-syntax.yml
garloff: Works for me. Feel free to push 1.4.6 to the branch.

garloff:
Feel free to push 1.4.6 to this branch and approve and merge it.
This is only the version used for the syntax/formatting check, of course.
The real version is the one referenced in main.tf of course.
We should not be too aggressive requiring the latest and greatest, as this pushes additional work on our users.
Unless we have good reasons for it, which seems to be the case here with `terraform_data`.

matofeder:
> Feel free to push 1.4.6 to this branch and approve and merge it. This is only the version used for the syntax/formatting check, of course. The real version is the one referenced in main.tf of course. We should not be too aggressive requiring the latest and greatest, as this pushes additional work on our users. Unless we have good reasons for it, which seems to be the case here with `terraform_data`.

I hope that our users won't be affected so much and they will understand the benefits such as security fixes and features that allow us to produce better SCS software for them. 

Besides [terraform_data](https://developer.hashicorp.com/terraform/language/resources/terraform-data) I will plan to use also another feature:  [optional-object-type-attributes](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#optional-object-type-attributes) in https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/432 which simplifies user interface a little bit.

matofeder (approved):
LGTM! 

/tested with TF version 1.4.6.

PR #428, opened 2023-06-17
Update to latest k8s and latest occm versions.

garloff:
We have k8s-v1.24.15, 1.25.11, 1.26.6 and 1.27.2 now. And OCCM-v1.25.6, 1.26.3 and 1.27.1.

The v1.27 versions are marked as experimental, until we have done more validation and/or official support from capo.


batistein (approved):

chess-knight (commented):
The images for the ccsi and occm are now in the *registry.k8s.io/provider-os* registry. So these workarounds need to be most likely changed [ccsi](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/bin/apply_cindercsi.sh#L60), [occm](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/bin/apply_openstack_integration.sh#L33).
Or add another line with the new registry, or they can be probably deleted as the upstream issue 2094 is resolved.

Edit: Or we can keep it as it is, so we will keep support for existing <=1.24.x tags and rely on the correct tags for new releases.
I don't know which one is better :smile:

garloff:
> The images for the ccsi and occm are now in the _registry.k8s.io/provider-os_ registry. So these workarounds need to be most likely changed [ccsi](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/bin/apply_cindercsi.sh#L60), [occm](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/bin/apply_openstack_integration.sh#L33). Or add another line with the new registry, or they can be probably deleted as the upstream issue 2094 is resolved.
> 
> Edit: Or we can keep it as it is, so we will keep support for existing <=1.24.x tags and rely on the correct tags for new releases. I don't know which one is better smile

So, the newer deployment manifests do
(a) reference images from registry.k8s.io/provider-os rather than docker.io/k8scloudprovider
      This means that the workaround becomes ineffective as the `sed` command will not match.
(b) do not require the workaround, as upstream issue 2094 has been resolved,

Do I understand this correctly?
The question then is:
1. If all new deployment manifests that reference the new registry also refer to versions where the correct container images are referenced (bug 2094 is fixed), then we don't need to do anything. Other than at some later point in time remove the workaround to improve maintainability of the code.
2. If there are new manifests even for older versions, then we may be able to drop the workaround completely right away.
3. If not all manifests that have the new registry also have the container versions fixed, then we'd need the second line to also fixup the versions there.

My preference would be 2., 1., 3.

chess-knight:
> So, the newer deployment manifests do
(a) reference images from registry.k8s.io/provider-os rather than docker.io/k8scloudprovider
This means that the workaround becomes ineffective as the sed command will not match.
(b) do not require the workaround, as upstream issue 2094 has been resolved,

> Do I understand this correctly?

a) Right. Starting from [v1.27.0](https://github.com/kubernetes/cloud-provider-openstack/releases/tag/v1.27.0) *registry.k8s.io/provider-os* is used. They also changed it from [v1.26.3](https://github.com/kubernetes/cloud-provider-openstack/releases/tag/v1.26.3) and [v1.25.6](https://github.com/kubernetes/cloud-provider-openstack/releases/tag/v1.25.6). But for the older versions, the command is still effective.
b) The bug 2094 was resolved like they modified [release-procedure.md](https://github.com/kubernetes/cloud-provider-openstack/blob/master/release-procedure.md) and there are correct image tags from the [v1.24.6](https://github.com/kubernetes/cloud-provider-openstack/releases/tag/v1.24.6), [v1.25.4](https://github.com/kubernetes/cloud-provider-openstack/releases/tag/v1.25.4) and [v1.26.1](https://github.com/kubernetes/cloud-provider-openstack/releases/tag/v1.26.1). For older versions, we should probably keep a workaround.

> The question then is:

> 1. If all new deployment manifests that reference the new registry also refer to versions where the correct container images are referenced (bug 2094 is fixed), then we don't need to do anything. Other than at some later point in time remove the workaround to improve maintainability of the code.

As I wrote above, it seems that this is correct.

> 2. If there are new manifests even for older versions, then we may be able to drop the workaround completely right away.

As I wrote above, new manifests are only for the few latest versions. Also in that bug issue, they wrote that "usually we backport up to N-3 older version"

> 3. If not all manifests that have the new registry also have the container versions fixed, then we'd need the second line to also fixup the versions there.

This is not the case. See question 1.

Therefore, in my opinion, we should go with 1.(we don't need to do anything)

garloff:
@set-reminder 2023-10-15 clean up workaround for [issue 2904](https://github.com/kubernetes/cloud-provider-openstack/issues/2094) in the main branch (after R5 has been released and maintained/v6 been branched off).

garloff:
Thanks for your extensive review, @chess-knight, really appreciated.

Issue #144, opened 2022-03-02
cilium: 4 sig-network tests fail (sonobuoy)


garloff:
Testing a capo cluster with k8s-1.21.9, I consistently get 4 failed tests:
Failed tests:
```
[sig-network] Services should have session affinity timeout work for service with type clusterIP [LinuxOnly] [Conformance]
[sig-network] Services should have session affinity work for service with type clusterIP [LinuxOnly] [Conformance]
[sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]
[sig-network] Services should be able to switch session affinity for service with type clusterIP [LinuxOnly] [Conformance]

```
To Do: Investigate:
- Are they dependent on k8s version?
- Or anything that we'd need to take care of in our capi/capo setup?

garloff:
This only happens with cilium.

garloff:
With k8s-1.24.4 and cilium v1.12.2, we are down to one failed test.
```
[sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly]
 [Conformance]                                                                                                                          
```

garloff:
Retesting with k8s-1.25.3 and cilium v1.12.2, we still have the 4 failed tests:
```
Failed tests:                                                         
[It] [sig-network] Services should have session affinity work for service with type clusterIP [LinuxOnly] [Conformance]                     
[It] [sig-network] Services should have session affinity timeout work for service with type clusterIP [LinuxOnly] [Conformance]             
[It] [sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]                                                                                                                             
[It] [sig-network] Services should be able to switch session affinity for service with type clusterIP [LinuxOnly] [Conformance]             
```

chess-knight:
> Retesting with k8s-1.25.3 and cilium v1.12.2, we still have the 4 failed tests:
> 
> ```
> Failed tests:                                                         
> [It] [sig-network] Services should have session affinity work for service with type clusterIP [LinuxOnly] [Conformance]                     
> [It] [sig-network] Services should have session affinity timeout work for service with type clusterIP [LinuxOnly] [Conformance]             
> [It] [sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]                                                                                                                             
[...]

PR #430, opened 2023-06-17
Feat/cilium 1.13.3

garloff:


terraform/variables.tf
```diff
@@ -166,7 +166,7 @@ variable "use_cilium" {
 variable "cilium_binaries" {
   description = "cilium and hubble CLI versions in the vA.B.C;vX.Y.Z format"
   type        = string
-  default     = "v0.13.1;v0.11.2"
+  default     = "v0.13.3;v0.11.6"
```
chess-knight, line 169:
```suggestion
  default     = "v0.13.2;v0.11.6"
```
v0.13.3 doesn't exist. v0.13.2 is latest

terraform/files/bin/create_cluster.sh
```diff
@@ -157,7 +157,7 @@ echo "# Deploy services (CNI, OCCM, CSI, Metrics, Cert-Manager, Flux2, Ingress)"
 MTU_VALUE=$(yq eval '.MTU_VALUE' $CCCFG)
 if test "$USE_CILIUM" = "true" -o "${USE_CILIUM:0:1}" = "v"; then
   # FIXME: Do we need to allow overriding MTU here as well?
-  CILIUM_VERSION="v1.13.0"
+  CILIUM_VERSION="v1.13.3"
```
chess-knight, line 160:
```suggestion
  CILIUM_VERSION="v1.13.4"
```
why not **v1.13.4**?

terraform/environments/environment-default.tfvars
```diff
@@ -10,7 +10,7 @@ ssh_username         = "<username_for_ssh>"	  # defaults to "ubuntu"
 clusterapi_version   = "<1.x.y>"		  # defaults to "1.3.5"
 capi_openstack_version = "<0.x.y>"		  # defaults to "0.7.1"
 image                = "<glance_image>"		  # defaults to "Ubuntu 22.04"
-cilium_binaries      = "<v0.aa.bb;v0.xx.yy>"  # defaults to "v0.13.1;v0.11.2"
+cilium_binaries      = "<v0.aa.bb;v0.xx.yy>"  # defaults to "v0.13.3;v0.11.6"
```
chess-knight, line 13:
```suggestion
cilium_binaries      = "<v0.aa.bb;v0.xx.yy>"  # defaults to "v0.13.2;v0.11.6"
```

doc/configuration.md
```diff
@@ -21,7 +21,7 @@ Parameters controlling the Cluster-API management server (capi management server
 | `ssh_username`           |                 | SCS        | `ubuntu`          | Name of the default user for the `image`                                      |
 | `clusterapi_version`     |                 | SCS        | `1.3.5`           | Version of the cluster-API incl. `clusterctl`                                 |
 | `capi_openstack_version` |                 | SCS        | `0.7.1`           | Version of the cluster-api-provider-openstack (needs to fit the CAPI version) |
-| `cilium_binaries`        |                 | SCS        | `v0.13.1;v0.11.2` | Versions of the cilium and hubble CLI in the vA.B.C;vX.Y.Z format             |
+| `cilium_binaries`        |                 | SCS        | `v0.13.3;v0.11.6` | Versions of the cilium and hubble CLI in the vA.B.C;vX.Y.Z format             |
```
chess-knight, line 24:
```suggestion
| `cilium_binaries`        |                 | SCS        | `v0.13.2;v0.11.6` | Versions of the cilium and hubble CLI in the vA.B.C;vX.Y.Z format             |
```

terraform/files/bin/create_cluster.sh
chess-knight: Ok I see, v1.13.4 seems to break IPsec

doc/configuration.md
garloff: Thanks for spotting this, @chess-knight !
Rather than accepting your 3 individual suggestions via github (resulting in 3 commits on the branch), I created one commit that fixes all three places.

chess-knight:
I run sonobuoy conformance tests(`--mode certified-conformance`) and I discovered that cilium has most likely a problem with host port functionality because the test **[It] [sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]**(source is probably [this](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/hostport.go)) fails.

Cilium host port is documented [here](https://docs.cilium.io/en/v1.13/installation/cni-chaining-portmap/).
I tried with Calico installation and it works, probably because the host port is in Calico enabled by [default](https://docs.tigera.io/calico/3.25/reference/configure-cni-plugins#cni-network-configuration-lists).

garloff:
I believe this failure has existed forever.
See issue #144.
We even mention it in the release notes.
It should not prevent the version upgrade, as it's no regression.

But maybe we should investigate more before we make cilium the default? #431

chess-knight:
I made some tests, and always only the HostPort test fails. I tested also the affected version(1.25.3) mentioned in #144 - commented [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/issues/144#issuecomment-1602226179)

garloff:
I agree -- so please approve and merge!

chess-knight (approved):

PR #432, opened 2023-06-19
Add option to use custom container registry in containerd

### LINKED ISSUE #417: Allow cluster operators to use custom container registry
As a cluster operator, I want to use a custom container registry that mirrors public registries e.g. to have a copy of container images under my own control or to avoid issues with [pull rate limit](https://docs.docker.com/docker-hub/download-rate-limit/) from docker.io registry. 

### Current state

SCS K8s bootstrap scripts install various k8s components and software, a few of them use docker.io public registry as a source.
The docker.io source is defined in their manifest files. Some of them are located in the k8s-cluster-api-provider repository, and some of them are downloaded on the fly from other git repositories. 

### Desired state

As was discussed in one SCS container team meeting we agreed that it is a good idea to use `registry.scs.community` as a mirror for `docker.io` and maybe also other container sources to avoid pull rate limit issues. 

### Open questions

1. Should we hard-code the SCS registry  `registry.scs.community` instead of docker.io in related manifest files which are located in the k8s-cluster-api-provider repository and also patch downloaded manifest files before they are applied?

2. Should we introduce variables that allow cluster operators to override docker.io occurrences from 1.? It means: keeping `docker.io` as a default and allowing cluster operators to change this default. 

3. Should we do the same (1. or 2.) for other registries like  `k8s.gcr.io` (i.e. not only docker.io)? 

4. User workload-related question: Should we set CRI (containerd) default registry to  `registry.scs.community`? or make this value also configurable? It means: if the user does not specify the container registry in his/her manifest file (e.g. image: nginx), the `registry.scs.community` will be used as a default registry instead of `docker.io`

Your feedback is welcome, @jschoone, @garloff, @chess-knight

matofeder:
This PR adds the following:
- Introduces `containerd_registry_files` variable that allows SCS KaaS users to
    pass containerd registry hosts config files that configure containerd
- Sets the default value of the `containerd_registry_files` variable (file `terraform/files/containerd/docker.io`)
    that instructs containerd to use `registry.scs.community` container registry instance as a public mirror of DockerHub
- Script `terraform/files/bin/configure_containerd.sh` that injects containerd registry host and cert files into cluster-template.yaml. This allows specifying extra files to be passed to user_data upon creation of cluster nodes
- Related docs page `doc/usage/containter-registry-configuration.md`
    
Closes #417


chess-knight (approved):
I tested it with an additional *registry.dnation.cloud* mirror and it works for me. I mirrored scs registry with this config(files/containerd/registry.scs.community):
```toml
server = "https://registry.scs.community"

[host."https://registry.dnation.cloud/v2/scs"]
    capabilities = ["pull"]
    override_path = true
```

matofeder:
Marked as a draft, because the TF version has been bumped to 1.4.6 in https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/426. This allows using some "new" TF features like `terraform_data` or `optional` arguments. 
These will be used after #426

matofeder:
> Marked as a draft, because the TF version has been bumped to 1.4.6 in #426. This allows using some "new" TF features like `terraform_data` or `optional` arguments. These will be used after #426

 TF scripts have been updated  as follows:
    - `Deprecated `null_resource` has been replaced by `terraform_data`
      resource
    - Optional Object Type Attributes have been used in `containerd_registry_files` variable

see c08e89c953bb9e7ccbecb096f5c62d9f3f19a3b5

@chess-knight please re-test/re-review PR, thx!

chess-knight (approved):
> please re-test/re-review PR, thx!

Ok I re-tested it and LGTM!

PR #427, opened 2023-06-17
Update CAPI to v1.3.8 and CAPO to 0.7.3.

garloff:
These are bugfix only changes.
CAPI-v1.3.8:
    clusterctl: Return error on infra cluster and control plane discovery (#8610)
    e2e: Adjust machinepool helper e2e timeout (#8755)
    e2e: test/e2e check for machines being ready after provisioning on Runtime SDK test (#8647)
    e2e: test/framework fix docker pod log collector (#8644)
    KCP: Allow machine rollout if cert reconcile fails (#8738)
    KCP: Prevent KCP to create many private keys for each reconcile (#8626)
    CAPD: change the haproxy entrypoint to prevent getting stopped immediately after start (#8741)
    CAPD: Delegate CAPD port selection to the container runtime (#8654)
CAPI-v1.3.7:
    clusterctl: Use local kustomize version in create-local-repository.py (#8436)
    Dependency: Update kindnetd and kindest/haproxy (#8471)
CAPI-v1.3.6:
    API: Implements Getter interface for IPAddressClaim object (#8380)
    test: Make it possible to run envtest-based integration tests from vscode (#8212)

CAPO-v0.7.3:
    Always filter cluster subnets by cluster network ID (#1573)
CAPO-v0.7.2:
    Backport Octavia OVN provider support to release-0.7 (#1529)
    Infer port network from subnet (#1526)

PR #434, opened 2023-06-22
Use Ubuntu 22.04 images for the latest k8s versions.

### LINKED ISSUE #409: Test Ubuntu 22.04 images
Since the subiquity bug should be fixed now, let's test the Ubuntu 22.04 CAPI images again.

As discussed @matofeder @chess-knight, I would be grateful if you could test them. Some groundwork has been done in branch [feat/use-ubu2204](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/tree/feat/use-ubu2204-images)

chess-knight:
This uses Ubuntu 22.04 starting from 1.25.11, 1.26.6 and 1.27.3 as discussed in #409.

These images are working Ubuntu 22.04 images also with a small qcow2 disk size ~4,35GB. Also thanks to @berendt for rebuilding them after the upstream merge.

Working images are also versions >= *v1.24.14, v1.25.10, v1.26.5, v1.27.1* but these images have big disk size (~11,6GB on average) and have some problems with converting to raw, see this [comment](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/issues/409#issuecomment-1571970466). So I decided not to use them for 2204.

Fixes #409


garloff (commented):
I guess this is something we should mention in the release notes.
Otherwise looks great to me.

chess-knight:
> I guess this is something we should mention in the release notes. Otherwise looks great to me.

Thank you, I added release notes in d6892521922027f00903e7466fdfa434c6469d14.

garloff (approved):
Looks good to me now, great.
Sidenote:
* I assume OSISM does not build all node images with both 20.04 and 22.04 going forward. This means we need to cherry-pick this patch into 5.x and 5.0.x (and thus also #427, which we should do anyway.)

berendt:
> Looks good to me now, great. Sidenote:
> 
> * I assume OSISM does not build all node images with both 20.04 and 22.04 going forward. This means we need to cherry-pick this patch into 5.x and 5.0.x (and thus also [Update CAPI to v1.3.8 and CAPO to 0.7.3. #427](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/427), which we should do anyway.)

Correct. We disabled all Ubuntu 20.04 builds. I think it makes no sense to continue to build Ubuntu 20.04 images when Ubuntu 22.04 is now working like expected.

terraform/files/bin/upload_capi_image.sh
```diff
@@ -4,19 +4,30 @@
 
 ~/bin/fixup_k8s_version.sh $CCCFG
 KUBERNETES_VERSION=$(yq eval '.KUBERNETES_VERSION' $CCCFG)
+. ~/bin/parse_k8s_version.inc
 #UBU_IMG_NM=ubuntu-capi-image-$KUBERNETES_VERSION
 UBU_IMG_NM=$(yq eval '.OPENSTACK_IMAGE_NAME' $CCCFG)
```
matofeder, line 9: I see one possible edge case that we could consider:

`OPENSTACK_IMAGE_NAME` variable consists of a static string and `kubernetes_version`  variable, i.e. `ubuntu-capi-image-${kubernetes_version}`

IIUC the OpenStack image name for both supported Ubuntu versions (20.04 and 22.04) will be the same if the k8s version will not change e.g. for the default k8s v1.25.11 the OpenStack image name for both Ubuntu versions will be  `ubuntu-capi-image-v1.25.11`. 

IMO This may cause an unwanted behavior when the user OpenStack project already contains Ubuntu 20.04 image for the k8s version v1.25.11 or newer. Then the Ubuntu 22.04 image won't be uploaded to the user's OpenStack project as the image with the desired name is already there, see [upload_capi_image.sh#L20](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/bin/upload_capi_image.sh#L20)

Possible solutions:
1. Document the above and force users to remove "conflicting" images from their OS projects 
2. Introduce the Ubuntu version in  `OPENSTACK_IMAGE_NAME` variable, i.e. `ubuntu-${ubuntu_version}-capi-image-${kubernetes_version}`  (maybe consider applying this only for Ubuntu 22.04 and newer to avoid unwanted uploads of already uploaded 20.04 images)  (IMO: this looks good to me)
3. Replace OpenStack image if Ubuntu 22.04 and  k8s version v1.25.11 or newer are used

jschoone: Hi, @chess-knight, as discussed with @matofeder it is enough to do just document the change but as there's nothing expected to break users should not be forced to remove 20.04er images since they will get the 22.04 automatically when they update regularly.

matofeder (approved):

PR #438, opened 2023-06-23
Update kind to 0.18.0

chess-knight:
K8s version is now 1.26.3.

We cannot upgrade to 0.19 or 0.20 for now because capi is still [v1.3.x](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/427) and for kubernetes 1.27 capi [v1.4 is required](https://cluster-api.sigs.k8s.io/reference/versions.html).


matofeder (approved):
Lgtm!

PR #440, opened 2023-06-26
Update default ingress-nginx to 1.8.0

chess-knight:
See changelog https://github.com/kubernetes/ingress-nginx/releases.

IMO notable changes:
v1.7.0 removed support for k8s v1.23
v1.7.1 added support for k8s v1.27
v1.8.0 -  Alpine 3.18

I tested it with sonobuoy `--mode quick`.


terraform/files/bin/apply_nginx_ingress.sh
```diff
@@ -5,7 +5,7 @@ export KUBECONFIG=~/.kube/config
 KUBERNETES_VERSION=$(yq eval '.KUBERNETES_VERSION' $CCCFG)
 DEPLOY_NGINX_INGRESS=$(yq eval '.DEPLOY_NGINX_INGRESS' $CCCFG)
 if test "$DEPLOY_NGINX_INGRESS" = "true"; then
-	if test "${KUBERNETES_VERSION:0:4}" = "v1.1"; then NGINX_VERSION="v1.0.2"; else NGINX_VERSION="v1.6.4"; fi
+	if test "${KUBERNETES_VERSION:0:4}" = "v1.1"; then NGINX_VERSION="v1.0.2"; else NGINX_VERSION="v1.8.0"; fi
```
matofeder, line 8: IMO This should be adjusted to ensure that the nginx ingress controller works with any SCS-supported k8s version.
As you already mentioned, nginx ingress controller  v1.7.0 dropped support for k8s v1.23, hence this should be considered in the above expression.

matofeder:
/tested with k8s v1.25.11 

- works as expected with the following demo apps with ingress :
<details>
  <summary>demo</summary>
  
    kind: Pod
    apiVersion: v1
    metadata:
      name: apple-app
      labels:
        app: apple
    spec:
      containers:
        - name: apple-app
          image: hashicorp/http-echo
          args:
            - "-text=apple"
    
    ---
    
    kind: Service
    apiVersion: v1
    metadata:
      name: apple-service
    spec:
      selector:
        app: apple
      ports:
        - port: 5678 # Default port for image
    ---
    kind: Pod
    apiVersion: v1
    metadata:
      name: banana-app
      labels:
        app: banana
    spec:
      containers:
        - name: banana-app
          image: hashicorp/http-echo
          args:
            - "-text=banana"
    
    ---
    
    kind: Service
    apiVersion: v1
    metadata:
      name: banana-service
    spec:
      selector:
        app: banana
      ports:
        - port: 5678 # Default port for image
    ---
    apiVersion: networking.k8s.io/v1
    kind: Ingress
    metadata:
      name: example-ingress
      annotations:
        ingress.kubernetes.io/rewrite-target: /
    spec:
      ingressClassName: nginx
      rules:
      - http:
          paths:
            - path: /apple
              pathType: Prefix
              backend:
                service:
                  name: apple-service
                  port:
                    number: 5678
            - path: /banana
              pathType: Prefix
              backend:
                service:
                  name: banana-service
                  port:
                    number: 5678

</details>

as I did not find any suitable sonobuoy tests for testing the nginx-ingress controller

terraform/files/bin/apply_nginx_ingress.sh
chess-knight: What is the SCS-supported k8s version? If we follow Kubernetes release cycle we are fine here, because 1.23 is [not supported](https://kubernetes.io/releases/patch-releases/#non-active-branch-history)

matofeder: ### What is the SCS-supported k8s version? 
- Docs mentioned that `As of 9/2022 our tests cover 1.21.latest ... 1.25.latest.`, see [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/doc/configuration.md#supported-k8s-versions) 
    -  IMO this should be updated and the above should match the R4 release notes (see below) 
 
- R4 release nodes mentioned `1.22 -- 1.26`  and that we had tested earlier versions (down to 1.18) successfully before, and we don't expect them to break, see [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/Release-Notes-R4.md#k8s-versions-122----126)

Based on the above docs I would say that **R4 supports k8s since v 1.22.**
  
-  SCS capi source code is able to deploy k8s from v1.18.20 to v1.27.2 as is defined [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/bin/openstack-kube-versions.inc#L6)

- In addition, there is an above expression and docs record that SCS code deploys ingress `v1.0.2` for k8s <= 1.19

For sure the expression and ingress-related docs were not maintained properly and we do not have a working ingress-controller for specific k8s versions (not-tested) see  [ingress-nginx#supported-versions-table](https://github.com/kubernetes/ingress-nginx#supported-versions-table).


### Proposals

Versioning related (! This could be discussed in a separate issue as it is not directly related to this issue): 
1. Bump k8s versions mentioned in [configuration.md#supported-k8s-versions](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/doc/configuration.md#supported-k8s-versions) as follows:   `As of R4, our tests cover 1.22.latest ... 1.26.latest ...`
    - This will match current release R4 notes

2. I would like to propose to follow [k8s release cycle](https://kubernetes.io/releases/patch-releases/#non-active-branch-history) for **R5** and  define k8s support for R5 as follows: ` 1.25.0 ... 1.27.latest`  (as the end of life for 1.24 is on 2023-07-28, which is before R5 I guess)  

Ingress related:

1. I see two possible solutions:
   - 1a) Provide full solution and adjust the above expression (`if test "${KUBERNETES_VERSION:0:4}" = "v1.1"; ...`) and related docs to match the [ingress-nginx#supported-versions-table](https://github.com/kubernetes/ingress-nginx#supported-versions-table), i.e. our source code will follow the ingress-k8s compatibility matrix
   - 1b) Remove partial solution, i.e. ingress `v1.0.2` for k8s <= 1.19, and document that we support ingress deployment only for k8s >=1.24 


@garloff, @jschoone your thoughts, and comments on this are welcomed

chess-knight: Additionally [here](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0210-v1-k8s-new-version-policy.md) is SCS standard about k8s version policy

jschoone: Hi @chess-knight and @matofeder, this is actually all a bit contradictory.
> Docs mentioned that As of 9/2022 our tests cover 1.21.latest ... 1.25.latest., see [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/doc/configuration.md#supported-k8s-versions)
Yes, that should be updated. Could you take care of it?

To the supported versions: As @chess-knight already mentioned that even v1.23 is EOL and also refers to our own standard we are forced to violate this if we still provide versions <=1.23 and there is a CVE for them because we simply can't fix them.

So I'm also fine with a combination `1a` and `1c` of @matofeder's solutions.

chess-knight: See 57a077dc6f187769b611cc025234d5fae7ece161

matofeder (approved):
Lgtm!

PR #446, opened 2023-06-28
Match supported k8s versions with R4

chess-knight:
As discussed in https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/440#discussion_r1243319562


matofeder (approved):
Lgtm!

PR #456, opened 2023-07-07
Add check to Makefile for clouds.yaml presence

fdobrovolny:


terraform/Makefile
```diff
@@ -60,7 +60,16 @@ dry-run: init
...
 mycloud: environments/environment-$(ENVIRONMENT).tfvars
-	@$(YQ) '.clouds."$(CLOUD)"' $(YQIN) < <(cat ./clouds.yaml ~/.config/openstack/clouds.yaml /etc/openstack/clouds.yaml 2>/dev/null) > mycloud.$(CLOUD).yaml
+	@if [ -f "clouds.yaml" ]; then \
+        $(YQ) '.clouds."$(CLOUD)"' $(YQIN) < clouds.yaml > mycloud.$(CLOUD).yaml; \
+    elif [ -f "~/.config/openstack/clouds.yaml" ]; then \
+        $(YQ) '.clouds."$(CLOUD)"' $(YQIN) < ~/.config/openstack/clouds.yaml > mycloud.$(CLOUD).yaml; \
+    elif [ -f "/etc/openstack/clouds.yaml" ]; then \
+        $(YQ) '.clouds."$(CLOUD)"' $(YQIN) < /etc/openstack/clouds.yaml > mycloud.$(CLOUD).yaml; \
+    else \
+        echo "Error: clouds.yaml file not found in any location."; \
```
chess-knight, line 70: Lgtm, but I don't know why my make autocompletion breaks with this line:
```bash
$ make m # [tab][tab]
echo    "Error
```

fdobrovolny: :thinking: I managed to reproduce it locally. I will try to look into it. The autocompletion works in general unless one tries to make autocomplete on "m"

matofeder (changes requested):
The second condition does not work as expected, because of quoted tilde. Quoting of `~` (or, disables its expansion, see e.g. [here](https://stackoverflow.com/questions/67117575/how-to-use-inside-a-double-quote-path-reference)

matofeder (approved):
LGTM!

chess-knight (approved):
LGTM

PR #474, opened 2023-07-25
Gets variable username from <env>.tfvars

mxmxchere:

Currently if $USERNAME is set in the environment that calls make, this value is used instead of 
the value supplied in environments/environment-<env>.tfvars. I am unsure whether this is intended
or unintended. However with this version the value in the *.tfvars is used to connect in the various
make targets

### DIFF
--- a/terraform/Makefile
+++ b/terraform/Makefile
@@ -8,7 +8,6 @@ SHELL=/bin/bash
 SED ?= sed
 #ENVIRONMENT = gx-bc
 OPENSTACK ?= openstack
-USERNAME ?= ubuntu
 CONSOLE = capi-mgmtcluster
 #TESTCLUSTER ?= testcluster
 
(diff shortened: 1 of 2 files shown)


terraform/Makefile
```diff
@@ -19,6 +18,7 @@ ifeq ($(NEED_OSCLOUD),1)
 else
   ENVIRONMENT ?= $(OS_CLOUD)
 endif
+USERNAME=$(shell ( grep '^ssh_username' environments/environment-$(ENVIRONMENT).tfvars || echo ubuntu ) | $(SED) 's@^ssh_username[^=]*= *"*\([^"]*\).*$$@\1@' )
```
matofeder, line 21: Thx for providing this patch. IMO it was unintended to not use `environment-.tfvars` ssh_username variable as a source for `$USERNAME` here. 

Would be great to adjust the relevant doc as well. I found only one place: https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/doc/make-reference.md#make-ssh

mxmxchere: ok, good hint i will change that as well

matofeder (approved):
LGTM! 
/tested

PR #489, opened 2023-08-08
Pass conformance tests with cilium

### LINKED ISSUE #144: cilium: 4 sig-network tests fail (sonobuoy)

chess-knight:
Fixes #144


mxmxchere (approved):
Looks very good to me as we discussed in #144

PR #481, opened 2023-08-03
Separate k8s clusters by namespace #228

fdobrovolny:
#228

PR #493, opened 2023-08-10
Ensure kubeconfig for cluster namespaces

chess-knight:
Fix also set-context command introduced in #481


fdobrovolny:
Just for context, this issue arises when `KUBECONFIG` is set to a workload cluster then a new namespace is created there.

PR #494, opened 2023-08-13
Fix delete_cluster.sh and others to handle per cluster ns on os #492

### LINKED ISSUE #492: Resources are not cleaned up
#481 introduced separation of k8s clusters by namespace.
Clusters are now in a separate namespace(with the same name as the cluster itself).
It seems that this causes problems with all clean-up commands(`delete_cluster.sh`, `make fullclean` ...)  and maybe other scripts(`add_cluster-network.sh`, `enable-cilium-sg-kube.sh` ...) where the hardcoded value `default` is used.

For e.g. after using `make fullclean` command, network and router `k8s-clusterapi-cluster-testcluster-testcluster` were not deleted from the OpenStack project.

fdobrovolny:
CAPI generates OpenStack resource names based on the NS the cluster is in. 

This PR so far handles only scripts on the management node. Either I or @garloff will address this issue in full clean as well.

### DIFF
--- a/terraform/files/bin/add_cluster-network.sh
+++ b/terraform/files/bin/add_cluster-network.sh
@@ -25,7 +25,8 @@ findnewnic()
 MGMT="$PREFIX-mgmtcluster"
 # FIXME: We already know the name ($PREFIX-mgmtcluster)
 #MGMT=$(openstack server list --name "$PREFIX-mgmtcluster" -f value -c Name)
-openstack server add network $MGMT k8s-clusterapi-cluster-default-$CLUSTER_NAME || exit
+NETWORK=$(openstack network list -f value -c Name | grep "k8s-clusterapi-cluster-\(default-${CLUSTER_NAME}\|${CLUSTER_NAME}-${CLUSTER_NAME}\)")
+openstack server add network $MGMT $NETWORK || exit
 WAIT=0
 while test $WAIT -lt 30; do
 	findnewnic
--- a/terraform/files/bin/enable-cilium-sg-kube.sh
+++ b/terraform/files/bin/enable-cilium-sg-kube.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 . ~/bin/cccfg.inc
-SGS=$(openstack security group list -f value -c ID -c Name | grep "k8s-cluster-default-${CLUSTER_NAME}-secgroup-")
+SGS=$(openstack security group list -f value -c ID -c Name | grep "k8s-cluster-\(default-${CLUSTER_NAME}\|${CLUSTER_NAME}-${CLUSTER_NAME}\)-secgroup-")
 SG_WORKER=$(echo "$SGS" | grep worker | cut -d " " -f1)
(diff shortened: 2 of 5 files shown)


terraform/files/bin/add_cluster-network.sh
```diff
@@ -25,7 +25,8 @@ findnewnic()
 MGMT="$PREFIX-mgmtcluster"
 # FIXME: We already know the name ($PREFIX-mgmtcluster)
 #MGMT=$(openstack server list --name "$PREFIX-mgmtcluster" -f value -c Name)
-openstack server add network $MGMT k8s-clusterapi-cluster-default-$CLUSTER_NAME || exit
+NETWORK=$(openstack network list -f value -c Name | grep "k8s-clusterapi-cluster-\(default-${CLUSTER_NAME}\|${CLUSTER_NAME}-${CLUSTER_NAME}\)")
+openstack server add network $MGMT $NETWORK || exit
```
garloff, line 29: Wouldn't just
`openstack server add network $MGMT k8s-clusterapi-cluster-${CLUSTER_NAME}-${CLUSTER_NAME}` do the job?
Or are you concerned about an old management host with old clusters (that were created before we used namespaces) and then the mgmt host got updated?

terraform/files/bin/enable-cilium-sg-kube.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 . ~/bin/cccfg.inc
-SGS=$(openstack security group list -f value -c ID -c Name | grep "k8s-cluster-default-${CLUSTER_NAME}-secgroup-")
+SGS=$(openstack security group list -f value -c ID -c Name | grep "k8s-cluster-\(default-${CLUSTER_NAME}\|${CLUSTER_NAME}-${CLUSTER_NAME}\)-secgroup-")
```
garloff, line 3: Same as above: This is done to support both old (default namespace) and new (`CLUSTER_NAME` namespace) clusters?

garloff (approved):
Code looks good.
Thinking about it, the fact that both old (default ns) and new (cluster specific ns) clusters are supported by the scripts is really good. This makes life a lot better for users that upgrade from old to new as the scripts really handle the mixture without any hassle and we don't need to write long documentation or upgrade scripts for this one.
I also tested the code successfully.

garloff:
I think this should be merged.
Changing the naming of resources as requested in #495 is a second step and that second step might consist of several sub-steps (some of which may not happen if we look at the tradeoff b/w effort, maintainability and usefulness).

garloff:
I guess I will create another PR for the `terraform/cleanup/cleanup.sh` script which is being used by `make fullclean`?
Sidenote: As we address #495, this would become easier, but let's ensure we keep things working also during intermediate states.
Sidenote2: Did you intend to include `cleanup.sh` in this PR? My feeling is that two PRs are appropriate, but this is not a strong opinion.

fdobrovolny:
Yes, I was thinking about possibly supporting both old and new clusters.

Make sense. I initially thought of having both in one, but it makes more sense to do two.
I will mark this PR as done then.

Issue #495, opened 2023-08-14
Prefix all resources with PREFIX and CLUSTER_NAME


garloff:
As a cluster administrator, I want to be able to identify all IaaS resources that belong to a cluster, so I know what to save and what I can possibly safely delete when cleaning up.

Looking at the various resources at the OpenStack level, we do not consistently prefix everything with (per mgmthost) prefix and clustername. Here's what we find:

## MgmtServer:
* VM: $PREFIX-mgmtcluster (OK)
* Root Volume: Unnamed (Not Good)
* AppCred: $PREFIX-appcred (OK)
* Port: $PREFIX-port (OK)
* Network: $PREFIX-net (OK)
* Subnet: $PREFIX-subnet (OK)
* FloatingIP: Unnamed (but that can't be fixed -- we could assign a tag ... or at least pu a description)
* SGs: $PREFIX-allow-icmp/ssh, $PREFIX-mgmt (OK) 
* Router: $PREFIX-rtr (OK)
* Images: ubuntu-capi-image-v1.xx.x (We want this, as these should be shared for performance and cost)

## Cluster
* VMs: $PREFIX-$CLUSTER_NAME-md/control-plane (OK)
* Root Volumes: $PREFIX-$CLUSTER_NAME-*-root (OK)
* AppCred: $PREFIX-$CLUSTER_NAME-appcred (OK)
* Ports: $PREFIX-$CLUSTER_NAME-md/control-plane-* (OK)
* Networks: k8s-clusterapi-cluster-$CLUSTER_NAME-$CLUSTER_NAME (Not good: Lacks $PREFIX)
* Subnets: dito (Not good)
* SGs: $PREFIX-$CLUSTER_NAME-cilium (OK), k8s-cluster-$CLUSTER_NAME-$CLUSTER_NAME-secgroup-worker/controlplane (Not good: Lacks $PREFIX)
* Router: k8s-clusterapi-cluster-$CLUSTER_NAME-$CLUSTER_NAME (Not good: Lacks $PREFIX)
* Floating IP (kubeapi): Unnamed (we can't fix that. The description has the $CLUSTER_NAME, but not the $PREFIX, cloud also use a tag)
* Loadbalancers (kubeapi): k8s-clusterapi-cluster-$CLUSTER_NAME-$CLUSTER_NAME-kubeapi (not good: Lacks $PREFIX)
## Ingress
[...]

PR #507, opened 2023-08-17
Update management node root volume and FIP naming #495

fdobrovolny:
One part of #495 focuses on Terraform resources.

### DIFF
--- a/terraform/mgmtcluster.tf
+++ b/terraform/mgmtcluster.tf
@@ -16,8 +16,12 @@ data "openstack_networking_network_v2" "extnet" {
 
 # - management cluster -
 resource "openstack_networking_floatingip_v2" "mgmtcluster_floatingip" {
-  pool       = var.external != "" ? var.external : data.openstack_networking_network_v2.extnet.name
-  depends_on = [openstack_networking_router_interface_v2.router_interface]
+  pool        = var.external != "" ? var.external : data.openstack_networking_network_v2.extnet.name
+  depends_on  = [openstack_networking_router_interface_v2.router_interface]
+  description = "Floating IP for the ${var.prefix} management cluster node"
+  tags = [
+    "${var.prefix}-mgmtcluster", "k8s-cluster-api-mgmtcluster"
+  ]
 }
 
 resource "openstack_networking_port_v2" "mgmtcluster_port" {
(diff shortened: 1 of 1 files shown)


terraform/mgmtcluster.tf
```diff
@@ -54,14 +65,18 @@ resource "openstack_compute_instance_v2" "mgmtcluster_server" {
...
-    source_type           = "image"
-    volume_size           = 30
+    uuid                  = openstack_blockstorage_volume_v3.mgmtcluster_volume.id
+    source_type           = "volume"
     boot_index            = 0
     destination_type      = "volume"
     delete_on_termination = true
   }
 
+  lifecycle {
```
garloff, line 75: I had to read https://itnext.io/how-and-when-to-ignore-lifecycle-changes-in-terraform-ed5bfb46e7ae to understand why this makes sense ...

terraform/mgmtcluster.tf
```diff
@@ -54,14 +65,18 @@ resource "openstack_compute_instance_v2" "mgmtcluster_server" {
 
   network { port = openstack_networking_port_v2.mgmtcluster_port.id }
   block_device {
-    uuid                  = data.openstack_images_image_ids_v2.images.ids[0]
-    source_type           = "image"
-    volume_size           = 30
+    uuid                  = openstack_blockstorage_volume_v3.mgmtcluster_volume.id
+    source_type           = "volume"
```
garloff, line 69: I assume there is no way to name the volume if we use source_type = "image" here?

terraform/mgmtcluster.tf
```diff
@@ -45,6 +49,13 @@ data "openstack_images_image_ids_v2" "images" {
   sort = "updated_at:desc"
 }
 
+resource "openstack_blockstorage_volume_v3" "mgmtcluster_volume" {
+  size        = 30
+  name        = "${var.prefix}-mgmtcluster"
```
garloff, line 54: I guess I would have named it "${var.prefix}-mgmtcluster-root" but that is more a thing of taste.
(Yes I also suffix routers with -rtr or so ...)

garloff (approved):
Looks good to me though you might want to quickly check the comments.

terraform/mgmtcluster.tf
fdobrovolny: 😁 Like it does not need to be there.

I originally was trying to accomplish that the volume would get created only when running on a new instalation, but that was not possible as that was cycle of dependencies.

Terraform apply is run only in make create. So unless someone runs make create again or runs apply manually this should not be needed.

If they did run it without this the management node root volume would get destroyed along with all data. 

With this it will create a new volume however the old one will stay attached and existing.

So it is backwards compatible and relatively safe except of possible creation of a new unused volume when user does something they should not.

terraform/mgmtcluster.tf
fdobrovolny: The issue is not the source_type but for some reason inside the instance you can't name the volume. It can be only done when the instance is defined as separate resource.

terraform/mgmtcluster.tf
fdobrovolny: Well my personal taste is usually:

"<project>-<service>-<env>-<location>-<name>"

I did it mostly the same as the instance was, but I'm more then happy to comply here.

PR #508, opened 2023-08-17
Detect yq being wrapper around jq and handle.

garloff:
Previous check was based on version number which was unreliable.

There are two rather different versions of yq out there and we need to treat them differently.
(We can chose which one to use on the mgmt host by installing our chosen favorite but not so from the bootstrap machine where we call `make create` to create the mgmt host.)
- The version from mikefarah (lua based)
- A thin wrapper around jq

We change the detection mechanism to look for jq in the help text as opposed to using version numbers which are evolving and have bitten us before.


garloff:
Any reviewers?

chess-knight (approved):
LGTM!

PR #510, opened 2023-08-22
Check and fail if prefix is in use

DEiselt:


mxmxchere (approved):
yes, very good does what it is supposed to do, lgtm

PR #515, opened 2023-08-24
Require terraform version < 1.6.0

chess-knight:
Protect users from accidentially using unfree BSL licensed code
Users can still manually update this version constraint if they want


garloff (approved):
Thanks, looks good to me.

Issue #521, opened 2023-08-28
Test migration from R4 to R5


jschoone:
As a container administrator, I want to have a working toolset so that I can migrate my clusters from R4 to R5.

## Definition of Ready:

- [ ] User Story is small enough to be finished within one sprint
- [ ] User Story is clear and understood by the whole team
- [ ] Acceptance criteria are defined
- [ ] Acceptance criteria are clear and understood by the whole team

## Definition of Done:

- [ ] Successfully used [Migration script](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/bin/update-R4-to-R5.sh)
- [ ] All acceptance criteria are met
- [ ] Changes have been reviewed
- [ ] CI tests have run successfully
- [ ] Documentation has been updated
- [ ] Release Notes have been updated

fdobrovolny:
@NotTheEvilOne, I might be wrong, but I think you got confused. R4 != 4.0.0. You should start on the `maintained/v5.x` and then check out to `main`. But I made a branch for you from the tag `v5.0.0` -> `release/v5.0.0`

NotTheEvilOne:
> @NotTheEvilOne, I might be wrong, but I think you got confused. R4 != 4.0.0. You should start on the `maintained/v5.x` and then check out to `main`. But I made a branch for you from the tag `v5.0.0` -> `release/v5.0.0`

I guess you are right. Ignore my previous comment in this case :)

NotTheEvilOne:
With help from @fdobrovolny and the `500-jq-R5` branch I have been able to both use `update-R4-to-R5.sh` script and run `create_cluster.sh`. I'm facing issues with `habor-settings` as mentioned by @fdobrovolny as well. This results in:
```
apiservice.apiregistration.k8s.io/v1beta1.metrics.k8s.io created
/home/ubuntu/bin/apply_nginx_ingress.sh: line 4: /home/ubuntu/capi-testcluster/harbor-settings: No such file or directory
Deploy NGINX ingress v1.8.0 controller to capi-testcluster
```

The cluster seems to come up successfully. Still need to investigate if all pods (especially NGINX) start as well.

PR #525, opened 2023-08-29
Enhance documentation of the `update-R4-to-R5.sh` script #521

fdobrovolny:
One fix under the #521 issue.


NotTheEvilOne (approved):
This change clarifies steps required for #521 to update existing and create new clusters.

PR #526, opened 2023-08-29
Feat/remove att volumes

### LINKED ISSUE #520: complete `make fullclean` after PR #507
After #507, we have an additional root volume that we may need to clean up in `make fullclean`.
We also have the opportunity to identify FIP addresses due to the added tag in #507.

Plan:
* Option `--force-fip-cleanup` to remove all unassociated FIPs,
   otherwise only remove the FIPs that we could pinpoint to the cleaned capi prefix before
   plus the one that we can now find from the tag.
* Clean the root volume from the capi management host
* Option `--force-pvc-cleanup` to remove all persistent volumes (from k8s clusters)

garloff:
Fixes #520.

We can do better in volume deletion:
* All volumes attached to our cluster nodes (workers and control plane) should be deleted. These are root volumes and possibly PVCs
* In addition, still identify them by name (just in case the servers have already been deleted) -- and we have a name now, thanks to #507
* WIth new option `--force-pvc`, we delete all volumes starting with `pvc` in the name.

We can also do better in FIP deletion:
* We have a tag now for the FIP attached to the mgmtcluster, delete FIP identified by it
* With the new option `--force-fip`, we also delete FIPs that are DOWN

This should improve `make fullclean` to deal with more sitations in case of strange failures.
SO it's a more robust mech for our e2e zuul testing.

We also offer the new target `make forceclean` also using the two new force options. SHould not normally be needed.

### DIFF
--- a/doc/make-reference.md
+++ b/doc/make-reference.md
@@ -85,11 +85,15 @@ unfortunately, some error in the deployment may result in resources left around.
 
 ### Make fullclean
 
-``make fullclean`` uses a custom script (using the openstack CLI) to clean up everything
-while trying to not hit any resources not created by the CAPI or terraform.
+``make fullclean`` uses a custom script `cleanup/cleanup.sh` (using the openstack CLI) to clean up
+everything while trying to not hit any resources not created by the CAPI or terraform for
+clusters from this management host.
 It is the recommended way for doing cleanups if ``make clean`` fails. Watch out for leftover
 floating IP addresses and persistent volumes, as these can not be easily traced back to the
-Cluster API created resources and may thus be left.
+Cluster API created resources and may thus be left. There is also a ``make forceclean`` variant
+that hits unused floating IPs and all persistent volumes -- this is risky as there is no good
+way to tell which PVCs belong to us unless we find them attached to cluster nodes in which
+case we don't need the force options.
 
 ### Make purge
 
--- a/terraform/cleanup/cleanup.sh
+++ b/terraform/cleanup/cleanup.sh
@@ -203,15 +203,15 @@ while read LB FIP; do
 		if test -z "$DEBUG"; then $OPENSTACK floating ip delete $FID; fi
 	fi
 done < <(echo "$LBS")
+SRV=$(resourcelist server $CAPIPRE-$CLUSTER)
+SRVVOL=$(server_vols $SRV)
+if test -n "$DEBUG"; then echo "### Attached volumes to "${SRV}": $SRVVOL"; fi
+cleanup server $CAPIPRE-$CLUSTER
 if test -n "$NOCASCADE"; then
 	cleanup_list loadbalancer 1 "" "$LBS"
 else
 	cleanup_list loadbalancer 1 "--cascade" "$LBS"
 fi
-SRV=$(resourcelist server $CAPIPRE-$CLUSTER)
-SRVVOL=$(server_vols $SRV)
-if test -n "$DEBUG"; then echo "### Attached volumes to ${SRV}: $SRVVOL"; fi
-cleanup server $CAPIPRE-$CLUSTER
 cleanup port $CAPIPRE-$CLUSTER
 RTR=$(resourcelist router "$CAPIPRE2ALL")
 SUBNETS=$(resourcelist subnet "$CAPIPRE2ALL")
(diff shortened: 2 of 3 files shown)


garloff:
Maybe one additional thing to mention:
We collect volumes when the servers still exist. Due to removal on termination, the root volumes will have gone already when we later try to delete all volumes. This will result in an error message:
```
openstack volume delete 7a8fb2e0-8b85-471a-a523-954d777994af cd5f1f3d-c9c3-47eb-b3c1-5a565cdff9af 9a4ad35e-97e1-480c-95e5-660ac95cf21d 9
2282e07-d722-4017-aa49-e90148182a5b c6a53b62-b111-4850-aa7f-0f36d82b5f7e                                                                
Failed to delete volume with name or ID 'cd5f1f3d-c9c3-47eb-b3c1-5a565cdff9af': No volume with a name or ID of 'cd5f1f3d-c9c3-47eb-b3c1-
5a565cdff9af' exists.                                                                                                                   
Failed to delete volume with name or ID '9a4ad35e-97e1-480c-95e5-660ac95cf21d': No volume with a name or ID of '9a4ad35e-97e1-480c-95e5-
660ac95cf21d' exists.                                                                                                                   
Failed to delete volume with name or ID '92282e07-d722-4017-aa49-e90148182a5b': No volume with a name or ID of '92282e07-d722-4017-aa49-
e90148182a5b' exists.
Failed to delete volume with name or ID 'c6a53b62-b111-4850-aa7f-0f36d82b5f7e': No volume with a name or ID of 'c6a53b62-b111-4850-aa7f-
0f36d82b5f7e' exists.
4 of 5 volumes failed to delete.
```
The volumes are gone and the one that is not gone already (the pvc) will still be deleted.
(Yes, the commands are NOT implemented as atomic all-or-nothing in the openstack-cli.)

terraform/cleanup/cleanup.sh
```diff
@@ -97,13 +100,34 @@ cleanup()
...
+# Output a list of volume UUIDs that are attached
+server_vols()
+{
+	if test -z "$1"; then echo; return 1; fi
+	srchexpr="\\("
+	for arg in "$@"; do
+		srchexpr="$srchexpr$arg\\|"
+	done
+	srchexpr="${srchexpr%\\|}\\)"
+	#if test -n "$DEBUG"; then echo "#### Search for volumes $srchexpr"; fi
```
fdobrovolny, line 114: Should this be commented out?

doc/make-reference.md
```diff
@@ -85,11 +85,15 @@ unfortunately, some error in the deployment may result in resources left around.
...
+``make fullclean`` uses a custom script `cleanup/cleanup.sh` (using the openstack CLI) to clean up
+everything while trying to not hit any resources not created by the CAPI or terraform for
+clusters from this management host.
 It is the recommended way for doing cleanups if ``make clean`` fails. Watch out for leftover
 floating IP addresses and persistent volumes, as these can not be easily traced back to the
-Cluster API created resources and may thus be left.
+Cluster API created resources and may thus be left. There is also a ``make forceclean`` variant
+that hits unused floating IPs and all persistent volumes -- this is risky as there is no good
+way to tell which PVCs belong to us unless we find them attached to cluster nodes in which
+case we don't need the force options.
```
fdobrovolny, line 96: Could you add a whole entry for the `make forceclean` command in the docs?

fdobrovolny (approved):
LGTM, it would be nice to update the docs.

terraform/cleanup/cleanup.sh
garloff: It is commented out already ... so I may fail to understand your suggestion?

doc/make-reference.md
garloff: I wanted to avoid it, as I don't recommend using it.
You should really be fine with fullclean -- and if that really ends up leaving resources uncleaned, finding the somewhat hidden mention of forceclean would hopefully also make you read the disclaimer.

fdobrovolny: Make sense.

terraform/cleanup/cleanup.sh
fdobrovolny: I meant it more like if it is a mistake that it is commented. But does not seem like it.

fdobrovolny (approved):

PR #527, opened 2023-08-30
Fix harbor-settings no such file

chess-knight:
Fix the error mentioned in #521


fdobrovolny (approved):
It could put some boilerplate in the file but, LGTM.

PR #537, opened 2023-08-31
Fix cilium install

### LINKED ISSUE #514: cilium-operator pod in CrashLoopBackOff
```bash
$ k logs -n kube-system cilium-operator-b5695c56f-zt7nx
...
level=fatal msg="Failed to create gateway controller" error="failed to get API group resources: unable to retrieve the complete list of server APIs: gateway.networking.k8s.io/v1beta1: the server could not find the requested resource" subsys=gateway-controller
```
It looks like #503 doesn't work as expected.
I deployed the cluster with default settings but in the logs from creation I have:
```
terraform_data.mgmtcluster_bootstrap_files (remote-exec): # Deploy services (CNI, OCCM, CSI, Metrics, Cert-Manager, Flux2, Ingress)
terraform_data.mgmtcluster_bootstrap_files (remote-exec): ℹ️  Using Cilium version 1.14.0
terraform_data.mgmtcluster_bootstrap_files (remote-exec): 🔮 Auto-detected cluster name: democluster
terraform_data.mgmtcluster_bootstrap_files (remote-exec): 🔮 Auto-detected datapath mode: tunnel
terraform_data.mgmtcluster_bootstrap_files (remote-exec): 🔮 Auto-detected kube-proxy has not been installed
terraform_data.mgmtcluster_bootstrap_files (remote-exec): ℹ️  Cilium will fully replace all functionalities of kube-proxy
```
Probably these settings:
```
    --helm-set-string "kubeProxyReplacement=${DEPLOY_GATEWAY_API}" \
    --helm-set-string "gatewayAPI.enabled=${DEPLOY_GATEWAY_API}" \
```
[here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/files/bin/create_cluster.sh#L188) do not work correctly because of `Cilium will fully replace all functionalities of kube-proxy`.

chess-knight:
gatewayAPI.enabled is a boolean, not a string.
Fixes #514

### DIFF
--- a/terraform/files/bin/create_cluster.sh
+++ b/terraform/files/bin/create_cluster.sh
@@ -175,8 +175,8 @@ done
 echo "# Deploy services (CNI, OCCM, CSI, Metrics, Cert-Manager, Flux2, Ingress)"
 MTU_VALUE=$(yq eval '.MTU_VALUE' $CCCFG)
 if test "$USE_CILIUM" = "true" -o "${USE_CILIUM:0:1}" = "v"; then
-  DEPLOY_GATEWAY_API=$(yq eval '.DEPLOY_GATEWAY_API' $CCCFG)
-  if test "$DEPLOY_GATEWAY_API" = "true"; then
+  DEPLOY_GATEWAY_API=$(yq eval '.DEPLOY_GATEWAY_API == true' $CCCFG)
+  if test "${DEPLOY_GATEWAY_API}" = "true"; then
     KUBECONFIG=${KUBECONFIG_WORKLOADCLUSTER} kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.7.0/config/crd/standard/gateway.networking.k8s.io_gatewayclasses.yaml
     KUBECONFIG=${KUBECONFIG_WORKLOADCLUSTER} kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.7.0/config/crd/standard/gateway.networking.k8s.io_gateways.yaml
     KUBECONFIG=${KUBECONFIG_WORKLOADCLUSTER} kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.7.0/config/crd/standard/gateway.networking.k8s.io_httproutes.yaml
@@ -191,8 +191,8 @@ if test "$USE_CILIUM" = "true" -o "${USE_CILIUM:0:1}" = "v"; then
   POD_CIDR=$(yq eval '.POD_CIDR' $CCCFG)
   KUBECONFIG=${KUBECONFIG_WORKLOADCLUSTER} cilium install --version $CILIUM_VERSION \
     --helm-set-string "ipam.operator.clusterPoolIPv4PodCIDRList={${POD_CIDR}}" \
-    --helm-set-string "kubeProxyReplacement=${DEPLOY_GATEWAY_API}" \
-    --helm-set gatewayAPI.enabled="${DEPLOY_GATEWAY_API}" \
+    --helm-set kubeProxyReplacement=${DEPLOY_GATEWAY_API} \
+    --helm-set gatewayAPI.enabled=${DEPLOY_GATEWAY_API} \
     --helm-set cni.chainingMode=portmap \
     --helm-set sessionAffinity=true
   touch ~/$CLUSTER_NAME/deployed-manifests.d/.cilium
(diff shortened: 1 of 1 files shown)


NotTheEvilOne (approved):
I've added some changes for the issues mentioned in #514 as well, please double check if I kept your intention on this fix :) With the changes both `true`, `false` as well as `null` are handled correctly for me.

chess-knight:
I tried your changes, with default settings but I still see `Cilium will fully replace all functionalities of kube-proxy` and also e.g.:
```
$ kubectl -n kube-system get cm cilium-config -oyaml | grep replacement
  kube-proxy-replacement: strict
  kube-proxy-replacement-healthz-bind-address: ""
$ helm get values cilium -n kube-system | grep Replacement
kubeProxyReplacement: strict
```

NotTheEvilOne:
> I tried your changes, with default settings but I still see `Cilium will fully replace all functionalities of kube-proxy` and also e.g.:
> 
> ```
> $ kubectl -n kube-system get cm cilium-config -oyaml | grep replacement
>   kube-proxy-replacement: strict
>   kube-proxy-replacement-healthz-bind-address: ""
> $ helm get values cilium -n kube-system | grep Replacement
> kubeProxyReplacement: strict
> ```

That's weird. Cilium clearly says it's a bool [1]:
```
            $ cilium install |CHART_VERSION| \
                --set kubeProxyReplacement=true \
                --set-string extraConfig.enable-envoy-config=true
```

But for deprecated options to work you might be right that we should handle it as a string [2].

I would like to check first why Cilium prints `Auto-detected kube-proxy has not been installed` as this seems to be the root cause of the Cilium's default behavior.

[1] https://github.com/cilium/cilium/blob/0da3f7e29e9bfaa860d652a7be7598a556cf9647/Documentation/network/servicemesh/installation.rst#installation
[2] https://github.com/cilium/cilium/blob/0da3f7e29e9bfaa860d652a7be7598a556cf9647/daemon/cmd/status.go#L222

chess-knight:
> I would like to check first why Cilium prints Auto-detected kube-proxy has not been installed as this seems to be the root cause of the Cilium's default behavior.

This can be seen [here](https://github.com/cilium/cilium-cli/blob/main/install/autodetect.go#L205) - cilium-cli autodetect.

NotTheEvilOne:
> > I would like to check first why Cilium prints Auto-detected kube-proxy has not been installed as this seems to be the root cause of the Cilium's default behavior.
> 
> This can be seen [here](https://github.com/cilium/cilium-cli/blob/main/install/autodetect.go#L205) - cilium-cli autodetect.

Interesting, now I know where `strict` is set as well [1]. But is it intended that we don't setup kube-proxy at all? In that case Cilium has no other choice than replacing it's functionality beside what we set as an option.

[1] https://github.com/cilium/cilium-cli/blob/19a57012227e1e842db4a124c937a870c1c2e536/install/autodetect.go#L263

NotTheEvilOne:
Setting up a new cluster with default options deploys a kube-proxy as expected (it can be disabled thought). That means we run `cilium install [...]` too fast and break auto detection that way.

```
kube-system                         kube-proxy-glwtg                                                 1/1     Running   0          15m
```

NotTheEvilOne:
> > Setting up a new cluster with default options deploys a kube-proxy as expected (it can be disabled thought). That means we run `cilium install [...]` too fast and break auto detection that way.
> > ```
> > kube-system                         kube-proxy-glwtg                                                 1/1     Running   0          15m
> > ```
> 
> That's True, it is what I discovered also [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/issues/144#issuecomment-1655774046).
> 
> Maybe we should add something like `kubectl wait` for the kube-proxy daemonset as autodetect does here https://github.com/cilium/cilium-cli/blob/main/install/autodetect.go#L214.

Unfortunately `kubectl $KCONTEXT -n kube-system wait --for=condition=ready daemonset/kube-proxy` does not work as the daemonset might not even be deployed when we reach the line in `create_cluster.sh`. Looking for alternatives at the moment.

NotTheEvilOne (approved):
Please double check as well @chess-knight

chess-knight:
Ok, I see that you used `kubectl get` instead of `wait` which is nice, but why do we need to use `wait` for all kube-proxy pods then?
Also, shouldn't this waiting be done only when the cilium is true?

NotTheEvilOne (approved):
Both variants (`kubectl get` to ensure kube-proxy has been applied and `kubectl wait` to ensure the service is defined and reachable) are used because:
- Cilium checks if the kube-proxy daemonset is created. This does not mean that it's reachable or even it's service is already applied.
- Ensuring that both the pod is running and the service is reachable is more of a future proof measurement to not touch the code each time auto detection for Cilium (or calico or whatever CNI might be supported in the future) changes.
- Same is true for not only running the check if `USE_CILIUM` is set to `true`. Another version of calico or another CNI supported might expect basic k8s set-up to have been completed successfully as well before being applied.
- Waiting for all pods to be ready ensures we do not deploy a CNI on a broken node for whatever reason. At the time of writing normal behavior is that at the time we wait only one pod is deployed anyway. I have not seen `kubectl wait` to actually take effect.

The code can be optimized for sure. As we want the blocker to be fixed I would suggest to merge the PR at the moment and do optimization in a future sprint / separate issue.

Ok, I see that you used kubectl get instead of wait which is nice, but why do we need to use wait for all kube-proxy pods then?
Also, shouldn't this waiting be done only when the cilium is true?

chess-knight:
Ok, I agree with you @NotTheEvilOne. We should definitely merge this before R5!

> At the time of writing normal behavior is that at the time we wait only one pod is deployed anyway

I think it is because we have at that time only one control-plane without other nodes, and that's why the wait is only for one pod.

PR #538, opened 2023-08-31
Fix missing value auth-url required for auth plugin password

### LINKED ISSUE #533: Missing value auth-url required for auth plugin password
When I run some *make target* command I always get:
```
Missing value auth-url required for auth plugin password
```
It happens when I have **export ENVIRONMENT** instead of **export OS_CLOUD** in my local env.

It looks like #510 introduced it with `shell ($(OPENSTACK) server list ...)` [command](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/Makefile#L43), where [exporting](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/terraform/Makefile#L17) *OS_CLOUD* inside Makefile seems insufficient for it.

chess-knight:
Fixes #533


DEiselt (approved):
Good catch.

PR #544, opened 2023-09-04
Update default calico to 3.26.1

chess-knight:
See changelog https://github.com/projectcalico/calico/releases.


fdobrovolny (approved):
LGTM

Should the change in the deployed manifest be in any way reflected in the upgrade script?

chess-knight:
> Should the change in the deployed manifest be in any way reflected in the upgrade script?

As far as I know, the manifest is not use anywhere

DEiselt (approved):
LGTM

PR #550, opened 2023-09-06
Fix cleanup.sh deleting just the first cluster

fdobrovolny:
I found this bug now, where only the first cluster would get deleted.


garloff (approved):
/lgtm

Issue #484, opened 2023-08-04
Improve deletion of workload clusters


chess-knight:
Right now I can see 3 possibilities for how the clusters can be deleted:
1. `make purge` - it purges the whole OpenStack project and this make command:
    - has some hardcoded values, e.g. **capi-keypair**
    - doesn't delete all the local files as *clean* or *full clean* does
    - doesn't delete application credentials
2. `make fullclean` - uses custom script *cleanup/cleanup.sh* which tries to delete resources created only by CAPI or terraform. I suggest the following:
    - investigate if make command and custom script work as expected
    - investigate and document how this make command can be used when multiple workload clusters are created (using *create_cluster.sh* from mgmt server) and want to be deleted
    - [documentation](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/doc/make-reference.md#make-fullclean) says that there are some leftover floating IP addresses and persistent volumes - investigate if this is still the case and if it cannot be improved
        - e.g. at the end of the script I can see **Volumes from Cinder CSI left:** and **openstack volume list | grep 'pvc-'**, it is desired to keep these volumes?
3. `make clean` - does ssh to the capi mgmt server and runs `delete_cluster.sh` script in for loop through all workload clusters.  This script can be used also manually from the mgmt server to delete desired workload cluster. I have these findings:
    - it deletes workload pods only from the default namespace - why? and usually, these pods are managed by deployment or similar k8s workload objects and will be recreated after the deletion
    - it deletes ingress-nginx deployment
    - it deletes PVCs only in the default namespace - this can be a problem if the deleted pods are recreated and these PVCs are mounted by the pods because PVCs will be in a terminating state forever and the delete command hangs and the user needs to press **^C** to continue
    - I think that the rest of the script needs to be investigated if it works as expected and there are also some TODOs
 
Maybe there are some other scripts/findings/bugs which I don't know about and need to be fixed.

Also, in #445, Harbor with ec2 credentials and the swift container is created and not removed when the workload cluster is deleted. Therefore I suggest as part of this issue to make a decision, if this behaviour should be preserved, e.g. because also `make fullclean` keeps PVCs, or it should be included in the mentioned scripts.

DEiselt:
[...]

PR #551, opened 2023-09-06
Update delete_cluster.sh to more thoroughly delete all k8s resources

fdobrovolny:
Addresses #484 `make clean` part.

This should try to remove all PV without the possibility of them being recreated.
As well as any load balancers.
As well as harbour OS resources.

### DIFF
--- a/terraform/files/bin/delete_cluster.sh
+++ b/terraform/files/bin/delete_cluster.sh
@@ -136,5 +136,9 @@ if grep '^OLD_OPENSTACK_CLOUD:' $CCCFG >/dev/null 2>&1; then
   fi
 fi
 rm -rf ~/$CLUSTER_NAME
-echo "Deleting cluster $CLUSTER_NAME done."
+if test $RC = 0; then
+  echo "Deleting cluster $CLUSTER_NAME completed successfully."
+else
+  echo "Deleting cluster $CLUSTER_NAME likely incomplete."
+fi
 exit $RC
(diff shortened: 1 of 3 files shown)


garloff (approved):
OK, we are much more thorough now with deleting anything at the k8s layer.
Looks good to me!

terraform/files/bin/delete_cluster.sh
```diff
@@ -90,5 +135,6 @@ if grep '^OLD_OPENSTACK_CLOUD:' $CCCFG >/dev/null 2>&1; then
     echo "Please delete application credential $PREFIX-$CLUSTER_NAME-appcred once capo has cleaned everything up."
   fi
 fi
-# TODO: Clean up ~/$CLUSTER_NAME
+rm -rf ~/$CLUSTER_NAME
+echo "Deleting cluster $CLUSTER_NAME done."
```
garloff, line 139:
```suggestion
if test $RC = 0; then
  echo "Deleting cluster $CLUSTER_NAME completed successfully."
else
  echo "Deleting cluster $CLUSTER_NAME likely incomplete."
fi
```

garloff: Just to avoid reporting that something has been done and then returning a contradicting error code that most likely will be ignored.
This is a rather cosmetic thing though.

matofeder (approved):
lgtm

garloff:
I just tested it successfully on a cluster without PVs.
Will redo a test with PVs and then merge if nothing unexpected happens.

PR #552, opened 2023-09-07
Feat/limit kube apiserv mem

### LINKED ISSUE #516: memory limits for kube-apiserver
As an operator of k8s clusters, I want to avoid the kube-apiservers to eat all memory ending up killing etcd (or the entire node).

Adapted from random k8s tip from @horazont in today's community call.

Let's quickly check whether this is something that affects us and could be addressed easily.

garloff:
This limits both CPU and Mem usage to 75% of what the control plane node offers.
With our default sizing (SCS-2V-4), this means max 1.5 vCPUs and 2.875GiB of memory for kube-apiserver.
Goal is to avoid an overwhelmed kube-apiserver causing the control plane node (and especially etcd) to become completely dysfunctional.
Fixes #516.

### DIFF
--- a/Release-Notes-R5.md
+++ b/Release-Notes-R5.md
@@ -228,6 +228,14 @@ With this followup PRs:
 - [#477](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/477)
   Add migration steps for existing k8s clusters to adopt #432
 
+### kube-apiserver resource limits
+
+The kube-apiserver has memory and cpu resource limits set now such that it will not eat all
+CPU and RAM on the control plane nodes, creating trouble for etcd or the CNI controller running there.
+This will make the control-plane a bit more robust under extremely high load and cause it
+to fail more gracefully if finally overwhelmed.
+[#552](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/552)
+
 ## Minor improvements
 
 - [#413](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/413)
--- a/terraform/files/template/cluster-template.yaml
+++ b/terraform/files/template/cluster-template.yaml
@@ -236,7 +236,7 @@ spec:
         content: |
           #!/bin/bash
           MEM=$(free -m | grep '^Mem:' | awk '{print $2;}')
-          CPU=$(grep '^processor:' /proc/cpuinfo | wc -l)
+          CPU=$(grep '^processor[   ]' /proc/cpuinfo | wc -l)
           sed -i "/^ *requests:/i\      limits:\n        memory: $((10+3*$MEM/4))M\n        cpu: $((750*$CPU))m" /etc/kubernetes/manifests/kube-apiserver.yaml
     postKubeadmCommands:
       - if test "${ETCD_UNSAFE_FS}" = "true"; then mount -o remount,barrier=0,commit=20 /; sed -i 's@errors=remount-ro@errors=remount-ro,barrier=0,commit=20@' /etc/fstab; fi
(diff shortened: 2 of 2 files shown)


garloff:
Not yet perfect: `k --context testcl7-admin@testcl7 -n kube-system get pod kube-apiserver-capi5-testcl7-control-plane-genc01-gwsj2 -o yaml`
```yaml
    resources:                                                                    
      limits:                                                                     
        cpu: 1500m                
        memory: 2944M    
      requests:        
        cpu: 250m                                                                 
        memory: 2944M              
```
The limit seems to implicitly also serve as request, so we end up filling up the control node and potentially preventing other services from running.

garloff:
```yaml
    resources:
      limits:
        cpu: 1500m
        memory: 2944M
      requests:
        cpu: 250m
        memory: 512M
```
This looks better now ...

garloff:
If this is common enough to emerge as a problem, should we mention this improvement in the release notes?

garloff:
> I was also working on this but have not made it work so far. I pushed the changes here:
> 
> [fdobrovolny@4043cdf](https://github.com/fdobrovolny/scs-k8s-cluster-api-provider/commit/4043cdf5f3bbb19c61ba4d58966e88d519e6305b)

I see you wanted to make this configurable ...
I was taking an easier route, just providing sane defaults without any tuning options.
Rationale:
* Too many choices can be overwhelming (even if they always have good defaults, bad if they don't).
* I consider this a quality/robustness of implementation thing, not so much a tunable that should be touched.
* We will migrate to SCS KaaS v2 with cluster stack in a few months. And carrying forward all these options is some extra work. If some of it comes later (to ensure the quality of implementation surpasses the old solution in most areas), it's easier if it's hidden ...

garloff:
> From my testing, I was able to make apiserver to die from OOM using the `kue-burner`. However in the end it was consuming all the CPU so I like that we limit the CPU usage as well. I managed to kill it using 100 requests per second it took a few minutes to die.

It will die also with the mem and cpu limits I guess -- although the CPU limit may have a bit of a throttling effect that allows it to survive a bit longer. What the limits should do is that the rest of the system -- especially etcd, but possibly also CNI controller -- still have enough resources to work reasonably well until kube-apiserver gets OOM killed.

If you have time to test with `kube-burner` again, it would be interesting to see how the control plane nodes fares with the current setting and check if reserved mem and cpu capacity are reasonably well tuned.

fdobrovolny:
> If this is common enough to emerge as a problem, should we mention this improvement in the release notes?

I think it deserves to be mentioned in the new features part. Some users may already have this implemented manually.

> I see you wanted to make this configurable ...
I was taking an easier route, just providing sane defaults without any tuning options.
Rationale

I guess I do see the beauty in that 99% of users will never need to change it or even know that it's there and it just works. But I also ultimately like the fact that if I need to, I can change the cluster template and edit it without too much hassle.

> If you have time to test with kube-burner again, it would be interesting to see how the control plane nodes fares with the current setting and check if reserved mem and cpu capacity are reasonably well tuned.

I'll try to test it and report back.

garloff:
Feel free to improve the text. Or just merge if it's fine.

matofeder (approved):
/tested

Test result:
- Kubeapiserver running on the control plane node with flavor `SCS-2V:4:20` got the following resource limits:

```bash
$ k get pod -n kube-system kube-apiserver-gx-scs-k8s-gx-scs-k8s-control-plane-genc01-zvbn5  -o=jsonpath='{.spec.containers[0].resources.limits}' | jq
{
  "cpu": "1500m",
  "memory": "2944M"
}

```

The above matches the 75% of resources available on the underlying control plane node. 

/lgtm

PR #557, opened 2023-09-08
Node images was moved to REGIO.cloud

chess-knight:
See osism/k8s-capi-images/pull/202

Additionally, k8s v1.28.1 can be deployed as techpreview.

It probably shouldn't be merged before R5!


NotTheEvilOne (approved):

fdobrovolny (approved):

berendt:
Thanks for watching out for the Minio service. I hadn't thought of this repository when I made the change. 

It would be good if you could change the source itself before the release, as we would like to shut down the Minio service as soon as possible. We will make a note in our 6.0.0 release notes that the systems will be shut down after the release in one month and the new sources are to be used.

PR #560, opened 2023-09-11
Improve make purge

chess-knight:
Addresses #484 `make purge` part.


fdobrovolny (approved):
LGTM while you are on this file could you maybe remove line 12?

PR #583, opened 2023-09-24
Fix logic for non-influencing execution of scripts

NotTheEvilOne:
This PR changes `create_cluster.sh` and `delete_cluster.sh` to allow them to be executed concurrently without side effects.


terraform/files/bin/create_appcred.sh
```diff
@@ -3,19 +3,27 @@
...
+  setup_kubectl_context_workspace
+  set_workload_cluster_kubectl_namespace
+fi
+
 # If the cluster exists already and we don't have a private appcred, leave it alone
 if kubectl get cluster $CLUSTER_NAME >/dev/null 2>&1 && ! grep '^OLD_OPENSTACK_CLOUD' ~/$CLUSTER_NAME/clusterctl.yaml >/dev/null 2>&1; then
 	echo "#Warn: Old style cluster, disable new appcred handling"
 	exit 0
 fi
-if kubectl get cluster $CLUSTER_NAME --namespace default >/dev/null 2>&1 && ! grep '^OLD_OPENSTACK_CLOUD' ~/$CLUSTER_NAME/clusterctl.yaml >/dev/null 2>&1; then
```
fdobrovolny, line 18: This has to stay here. Otherwise, it would be the same as on line 14. It is meant to support clusters in the default namespace instead of the new per-cluster namespaces.

terraform/files/bin/create_cluster.sh
```diff
@@ -80,9 +63,6 @@ export OS_CLOUD=$(yq eval '.OPENSTACK_CLOUD' $CCCFG)
 # Ensure image is there
 wait_capi_image.sh "$1" || exit 1
 
-# Switch to capi mgmt cluster
-export KUBECONFIG=$HOME/.kube/config
-~/bin/mng_cluster_ns.inc
```
fdobrovolny, line 85: This was historically placed here to reset to a known state after calling the "external_scripts".

terraform/files/bin/wait_capi_image.sh
```diff
@@ -21,4 +21,3 @@ while test $ctr -lt 180; do
   sleep 10
 done
 if test $ctr -ge 180; then echo "TIMEOUT"; exit 2; fi
-
```
fdobrovolny, line 24: Why?

fdobrovolny (changes requested):
LGTM, good work.

terraform/files/bin/create_cluster.sh
NotTheEvilOne: Understood. As the `kubeconfig` file is duplicated I guess this is no longer needed.

NotTheEvilOne:
@fdobrovolny, please let me know if further changes are requested here :)

chess-knight (commented):
`kubectl describe cluster <CLUSTER_NAME>` doesn't work for me now. It works only when I add `-n <CLUSTER_NAME>`.

If it is desired, the documentation should be updated.

chess-knight (commented):
When I work with workload cluster, I usually do `export KUBECONFIG=testcluster/testcluster.yaml`.

When I run `delete_cluster.sh` with that KUBECONFIG, I am getting these errors:
```
error: no context exists with the name: "kind-kind"
error: the server doesn't have a resource type "cluster"
> Cluster testcluster does not exist, and new namespace creation is disabled.
Deleting cluster testcluster
Cordoning all nodes ...
.
.
.
service "metrics-server" deleted
error: no context exists with the name: "kind-kind"
error: the server doesn't have a resource type "cluster"
warning: this removed your active context, use "kubectl config use-context" to select a different one
deleted context testcluster-admin@testcluster from testcluster/testcluster.yaml
deleted user testcluster-admin from testcluster/testcluster.yaml
deleted cluster testcluster from testcluster/testcluster.yaml
No Subnet found for k8s-clusterapi-cluster--testcluster
error: the server doesn't have a resource type "cluster"
error: the server doesn't have a resource type "cluster"
error: resource name may not be empty
Removing application credential capi-testcluster-appcred ...
Deleting cluster testcluster completed successfully.
```

Is it expected?

chess-knight (commented):
The similar problem as in https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/583#pullrequestreview-1652509633 is with `create_cluster.sh`

NotTheEvilOne:
> When I work with workload cluster, I usually do `export KUBECONFIG=testcluster/testcluster.yaml`.
> 
> When I run `delete_cluster.sh` with that KUBECONFIG, I am getting these errors:
> 
> ```
> error: no context exists with the name: "kind-kind"
> error: the server doesn't have a resource type "cluster"
> > Cluster testcluster does not exist, and new namespace creation is disabled.
> Deleting cluster testcluster
> Cordoning all nodes ...
> .
> .
> .
> service "metrics-server" deleted
> error: no context exists with the name: "kind-kind"
> error: the server doesn't have a resource type "cluster"
> warning: this removed your active context, use "kubectl config use-context" to select a different one
> deleted context testcluster-admin@testcluster from testcluster/testcluster.yaml
> deleted user testcluster-admin from testcluster/testcluster.yaml
> deleted cluster testcluster from testcluster/testcluster.yaml
> No Subnet found for k8s-clusterapi-cluster--testcluster
> error: the server doesn't have a resource type "cluster"
> error: the server doesn't have a resource type "cluster"
> error: resource name may not be empty
> Removing application credential capi-testcluster-appcred ...
> Deleting cluster testcluster completed successfully.
> ```
> 
> Is it expected?
> 
> Also, probably `remove_cluster-network.sh` need to be also adjusted

It's not expected but is it a valid use case? Exporting a workload cluster KUBECONFIG before calling `delete_cluster.sh` should by definition fail because it can not reach the management cluster.

It was working in the past because KUBECONFIG was overridden without any checks in one of the script files. As a side effect KUBECONFIG was changed for the user's session at the end of the execution. Additionally `~/.kube/config` was hard-coded.

Now the question is:
- Should we hard-code the old KUBECONFIG path (currently this branch takes the user's set one as the base for execution).
- Or should we error out if the context is not set for the current KUBECONFIG file?

NotTheEvilOne:
> LGTM, good work.

@fdobrovolny I would highly appreciate your input on the latest changes and if you can approve (currently it's still set to "request changes" your review result. Thanks :)

fdobrovolny (approved):

fdobrovolny:
From technical standpoint I have no comments it's good work and LGTM. 

From philosophical stands point I'm not 100% sure that this is the way to do it and I hoped someone else would share their opinion as well.

PR #584, opened 2023-09-25
Add option to specify external net via ID

### LINKED ISSUE #572: Bug: Having multiple external networks in a project does not work
**Part 1: Terraform does not allow multiple external networks (`make create`)**
```
╷
│ Error: Your query returned no results. Please change your search criteria and try again.
│
│   with data.openstack_networking_network_v2.external,
│   on neutron.tf line 68, in data "openstack_networking_network_v2" "external":
│   68: data "openstack_networking_network_v2" "external" {
│
╵
make: *** [Makefile:87: create] Error 1
```
**Part 2: CAPO does not support name in `OPENSTACK_EXTERNAL_NETWORK_ID`**

Currently k8s-cluster-api-provider allows both (id/name) for `external ` according to the docs: https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/353c15c64118e15197a9367e6913e3c14196d078/doc/configuration.md?plain=1#L35C120-L35C120
(in the past an ID didn't even work but I have not tried again)

That results in the OPENSTACK_EXTERNAL_NETWORK_ID being an name which is not supported by CAPO
- https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/docs/book/src/development/development.md#support-for-clouds-with-multiple-external-networks
- https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/353c15c64118e15197a9367e6913e3c14196d078/terraform/files/template/clusterctl.yaml.tmpl#L69

So CAPO does external network auto detection everytime and multiple external networks are not supported.

chess-knight:
Fixes #572

This commit should resolve these problems:
OPENSTACK_EXTERNAL_NETWORK_ID was prefilled with external network name instead of ID what causes external net autodetection by CAPO - multiple ext nets cause error. Documentation says that 'external' can be name/id of external network which doesnt work - introduce external_id parameter. Autodetection of external network is improved - when multiple ext net exists, it can be specified via external/external_id

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -29,13 +29,13 @@ Parameters controlling the Cluster-API management server (capi management server
 
 Parameters controlling both management server creation and cluster creation:
 
-| environment         | clusterctl.yaml                 | provenance | default                              | meaning                                                                                                                      |
-|---------------------|---------------------------------|------------|--------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
-| `cloud_provider`    | `OPENSTACK_CLOUD`               | capo       |                                      | `OS_CLOUD` name in clouds.yaml                                                                                               |
-| `external`          | `OPENSTACK_EXTERNAL_NETWORK_ID` | capo       | `""`                                 | Name/ID of the external (public) OpenStack network, default "" uses the detected external network                            |
-| `dns_nameservers`   | `OPENSTACK_DNS_NAMESERVERS`     | capo       | `[ "5.1.66.255", "185.150.99.255" ]` | Array of nameservers for capi mgmt server and for cluster nodes, replace the FF MUC defaults with local servers if available |
-| `availability_zone` | `OPENSTACK_FAILURE_DOMAIN`      | capo       |                                      | Availability Zone(s) for the mgmt node / workload clusters                                                                   |
-| `kind_mtu`          | `MTU_VALUE`                     | SCS        | `0`                                  | MTU for the mgmt server; Calico is set 50 bytes smaller; 0 means autodetection                                               |
+| environment              | clusterctl.yaml                 | provenance | default                              | meaning                                                                                                                                  |
+|--------------------------|---------------------------------|------------|--------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| `cloud_provider`         | `OPENSTACK_CLOUD`               | capo       |                                      | `OS_CLOUD` name in clouds.yaml                                                                                                           |
+| `external`/`external_id` | `OPENSTACK_EXTERNAL_NETWORK_ID` | capo       | `""`                                 | Name/ID of the external (public) OpenStack network, default "" uses the detected external network (for clouds with one external network) |
+| `dns_nameservers`        | `OPENSTACK_DNS_NAMESERVERS`     | capo       | `[ "5.1.66.255", "185.150.99.255" ]` | Array of nameservers for capi mgmt server and for cluster nodes, replace the FF MUC defaults with local servers if available             |
+| `availability_zone`      | `OPENSTACK_FAILURE_DOMAIN`      | capo       |                                      | Availability Zone(s) for the mgmt node / workload clusters                                                                               |
+| `kind_mtu`               | `MTU_VALUE`                     | SCS        | `0`                                  | MTU for the mgmt server; Calico is set 50 bytes smaller; 0 means autodetection                                                           |
 
 ### Parameters clusters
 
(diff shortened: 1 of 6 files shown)


DEiselt (approved):

doc/configuration.md
```diff
@@ -29,13 +29,13 @@ Parameters controlling the Cluster-API management server (capi management server
...
+|---------------------|---------------------------------|------------|--------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `cloud_provider`    | `OPENSTACK_CLOUD`               | capo       |                                      | `OS_CLOUD` name in clouds.yaml                                                                                                                                                                    |
+| `external`          | `OPENSTACK_EXTERNAL_NETWORK_ID` | capo       | `""`                                 | Name of the external (public) OpenStack network, default "" uses the detected external network(for clouds with one external network); use `external_id` if you want to specify network ID instead |
```
fdobrovolny, line 35: This is a bit confusing, because in the env file, this accepts only name, but in the `clusterctl.yaml` I would expect given the name an ID there, and what I find in the default one generated is indeed ID, but in the docs, it says a name should be there.

Maybe something like this?

```suggestion
| `external`/`external_id`          | `OPENSTACK_EXTERNAL_NETWORK_ID` | capo       | `""`                                 | Name or ID of the external (public) OpenStack network, default "" uses the detected external network(for clouds with one external network); use `external_id` if you want to specify network ID in the env file instead |
```

fdobrovolny (approved):
LGTM

doc/configuration.md
chess-knight: Please see the newest changes, I simplified it a little bit.

fdobrovolny (approved):
Great, thank you for the update.

terraform/neutron.tf
```diff
@@ -65,14 +65,10 @@ resource "openstack_networking_subnet_v2" "subnet_mgmt" {
   }
 }
 
-data "openstack_networking_network_v2" "external" {
-  name = var.external != "" ? var.external : data.openstack_networking_network_v2.extnet.name
-}
-
```
matofeder, line 71: I like the idea of avoiding multiple reads for the same OS resource. 
But I also prefer to put data sources next to the resources that reference them or have a shared `data.tf` file where you can put shared data.  

Overall I do not have a strong preference here, it is just a suggestion on how you could potentially improve script readability.

doc/configuration.md
```diff
@@ -29,13 +29,13 @@ Parameters controlling the Cluster-API management server (capi management server
...
+| environment              | clusterctl.yaml                 | provenance | default                              | meaning                                                                                                                                  |
+|--------------------------|---------------------------------|------------|--------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| `cloud_provider`         | `OPENSTACK_CLOUD`               | capo       |                                      | `OS_CLOUD` name in clouds.yaml                                                                                                           |
+| `external`/`external_id` | `OPENSTACK_EXTERNAL_NETWORK_ID` | capo       | `""`                                 | Name/ID of the external (public) OpenStack network, default "" uses the detected external network (for clouds with one external network) |
```
matofeder, line 35: We could also mention somewhere in the doc what will happen if the cloud has more then 1 external network available and the user does not specify the external/external_id parameter.

terraform/variables.tf
```diff
@@ -39,12 +39,19 @@ variable "availability_zone" {
...
   # default   = data.openstack_networking_network_v2.extnet.name
 }
 
+variable "external_id" {
+  description = "external/public network ID for access"
+  type        = string
+  default     = ""
+  # default   = data.openstack_networking_network_v2.extnet.id
+}
+
```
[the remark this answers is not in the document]
Nils98Ar: So terraform can also find the network by `id`?

If we want to support specifying the network by `name` for backwards compatibility I believe that we need a second variable with `id` for CAPO or we need to compute the `id` for CAPO.

Other options would be
- only support specifying the external network by `id` (no longer by `name`)
- don't support multiple external networks, remove the `external` variable and completely rely on auto-detection in terraform and CAPO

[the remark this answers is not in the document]
matofeder: If we want to support specifying the network by name for backwards compatibility I believe that we need a second variable with id for CAPO or we need to compute the id for CAPO

true, a smooth backward compatibility is a good reason for the second variable with ID

terraform/neutron.tf
chess-knight: See b15bb701f0c19e94504f2971aa19dd92c1aebde2

doc/configuration.md
chess-knight: See b15bb701f0c19e94504f2971aa19dd92c1aebde2

matofeder (approved):
/lgtm

PR #585, opened 2023-09-25
Revert "Remove unnecessary keystoneauth patch (#580)"

chess-knight:
This reverts commit d9fd9050ea69f916487b97784fa5206fd42e9e68.

This patch is still needed for the Ubuntu 20.04.

I found that Ubuntu 20.04 is used in some environments.

Without this patch, OpenStack commands end up with `Not Found (HTTP 404)` error.


DEiselt (approved):

NotTheEvilOne (approved):

matofeder (approved):
/lgtm

PR #596, opened 2023-09-27
Initial renovate config and renovate config for capi and capo

mxmxchere:

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,15 +1,16 @@
 {
-  "extends": [
-    "config:base"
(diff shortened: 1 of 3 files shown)


mxmxchere:
This also removed old renovate configuration, but since no instance was acting upon that config i removed it. It detected some terraform dependency and detected kustomization-dependencies (harbor). We can add it later again. For the beginning i want to start with a small setup.

DEiselt (approved):
LGTM as we tested this already

PR #600, opened 2023-09-27
Add ClusterClass

chess-knight:


terraform/files/template/cluster-template.yaml
```diff
@@ -11,268 +51,283 @@ spec:
...
+            name: '{{ local_hostname }}'
+            kubeletExtraArgs:
+              cloud-provider: external
+              provider-id: openstack:///'{{ instance_id }}'
+        files:
+          - path: /root/etcd-defrag.sh
+            owner: "root:root"
+            permissions: "0755"
+            content: |
+              #!/bin/bash
```
garloff, line 146: There seem to be a lot of unnecessary (whitespace ...) changes to the scripts.
I'd prefer them not to be included, as it complicates the review process and makes searching for issues with git annotate / git blame much harder.
If you think that this is valuable cleanup, please submit it in a separate PR.

chess-knight: Those are not whitespaces, but I moved the whole .spec of to .spec.template.spec so 4 spaces to the right. It is because we need `KubeadmControlPlaneTemplate` instead of `KubeadmControlPlane`.

fdobrovolny:
I think it would be great to separate cluster classes and current implementation as to retain the ability to update versions and do maintenance in existing clusters.

chess-knight:
This PR is breaking change! And in my eyes, we have two possibilities for how to handle it.

1. As @fdobrovolny suggested, have both options for how to deploy clusters (with/without ClusterClass). I.e. new clusters will be created using the ClusterClass feature and old clusters will still be using old cluster templates. But, this means, that we will need to maintain and support both approaches.
2. Introduce this breaking change without another option. I.e. new clusters will be created using the ClusterClass feature and old clusters **will need to be migrated**. I have found this nice [blog](https://eng.d2iq.com/blog/adopting-existing-clusters-to-use-clusterclass/) which shows how to adopt existing clusters to use ClusterClass (It is only for CAPD but hopefully it works also with CAPO).

IMHO n. 1 is not a good approach, because of maintainability. Also, this PR changes some core logic in scripts so it needs to be changed, and for the future, we will have to think about it constantly to support both approaches.

I prefer approach n. 2. This approach is also aligned with a thought of migration to ClusterStacks (our reference implementation v2). In this case, we will wait for the correct migration scripts and then we can merge this one.

Therefore, I suggest postponing this PR and working on migration scripts which will likely work as I mentioned in approach n. 2.

chess-knight:
Based on discussions with @jschoone, we will try option n. 2. I mark this PR as a Draft for now and continue work on migration scrips.

chess-knight:
> Based on discussions with @jschoone, we will try option n. 2. I mark this PR as a Draft for now and continue work on migration scrips.

I added a migration script in 2bd8c3a108b72965aa266141f072bdfd00e82025 so now it can be reviewed.

jschoone (approved):
The migration script works like a charm.
As we discussed we currently have an anti-pattern since every clusters has its own ClusterClass even if they look the same.
But since the migration is not a must for now, we will tackle this in another issue.

DEiselt (approved):

PR #608, opened 2023-10-06
Update terraform cache directory for providers

chess-knight:
The bug was, that terraform init was called all the time.
`.terraform/plugins` --> `.terraform/providers`

Also, prettify Makefile.


fdobrovolny (approved):
LGTM

However 🤔 I'm thinking if it would not be worth it to add `make tfinit` to give the user ability to manually trigger it for update. However most likely no update will ever come 😢

terraform/Makefile
```diff
@@ -4,7 +4,7 @@
 #ENVIRONMENT = gx-betacloud
 #ENVIRONMENT ?= gx-citycloud
 #ENVIRONMENT = gx-scs
-SHELL=/bin/bash
+SHELL = /bin/bash
```
matofeder, line 7: IMO it is not a good practice to address formatting issues in a pull request intended for a bug fix. I personally prefer having a separate pull request dedicated specifically to formatting changes.

matofeder (approved):
LGTM

PR #620, opened 2023-10-17
http_proxy support (part one: mgmt host)

mxmxchere:
It seems i have to make this draft-PR otherwise i cannot link my branch to the issue...


mxmxchere:
Because the CI is a bit picky at the moment: I ran e2e-quick-test once with http_proxy set and once without, both passed.

terraform/mgmtcluster.tf
```diff
@@ -99,6 +111,35 @@ write_files:
     owner: root:root
     path: /etc/needrestart/conf.d/needrestart.conf
     permissions: '0644'
+  - content: |
```
NotTheEvilOne, line 114: Just as a side note. These lines would be more readable if each list entry would start with the path instead of the content.

terraform/mgmtcluster.tf
```diff
@@ -111,10 +152,16 @@ runcmd:
   - sysctl -w -p /etc/sysctl.d/90-conntrack_max.conf
   - mkdir /etc/docker
   - HOME=/home/${var.ssh_username} /tmp/get_mtu.sh
-  - mv /tmp/daemon.json /etc/docker/daemon.json
+  - cp /tmp/daemon.json /etc/docker/daemon.json
```
NotTheEvilOne, line 155: Could you please describe why the move command needed to be changed here?

NotTheEvilOne (commented):
In general looks good to me. Shouldn't "HTTP Proxy" be written as "HTTP proxy"? I would appreciate if the `proxy.sh`  could be included less times.

terraform/mgmtcluster.tf
mxmxchere: i think that was for debugging purposes only, changed it back

terraform/mgmtcluster.tf
mxmxchere: thanks, i included that now

mxmxchere:
Testing check-quick without proxy passed.

NotTheEvilOne (approved):

Issue #418, opened 2023-05-12
Make use of HTTP and HTTPS proxy configurable


jschoone:
As an SCS Operator, I want to be able to configure the use of a corporate proxy in my Kubernetes clusters created by ClusterAPI.
Specifically I want to be able to specify the proxy settings in the ClusterAPI configuration and have these settings propagated to all nodes in the cluster during provisioning.

- [x] setup test proxy
- [x] introduce new config variable http_proxy

**Part1: mgmt host** (#620 )
- [x] /etc/profile mgmt host
- [x] apt for cloud-init
- [x] provisioning scripts
- [x] docker client mgmt host
- [x] docker daemon mgmt host

**Part2: worker and controlplane nodes** (#642)
- [x] ~~/etc/profile worker?~~
- [x] containerd worker
- [x] provisioning snippets (apt,curl...)
- [x] ~~kubelet worker~~

**Part3: Exceptions via `no_proxy`**
- [ ] create sensible defaults via pod,service and node cidr + k8s svc domain
- [ ] make additional entries configurabale via a new setting (document behaviour of both the default setting and additional entries)
- [ ] trickle down the settings to all places where http_proxy settings have been set

**Part4: Container workloads?**
- [ ] set env vars via mutatingwebhook

For now, i crossed /etc/profile from the list for the k8s-nodes. As /etc/profile is mostly used for login-shells. cluster-api managed-nodes are either created and provisioned automatically or deleted. Manual configuration via a login-shell (that would need proxy settings to be present in /etc/profile) is therefore unnecessary.

mxmxchere:
Status Update: the current state in https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/620 is enough to successfully deploy a testcluster with proxy configured on mgmt and kind and all containers running on the mgmt host. Next, i will try to trickle down the proxy settings into the created cluster.

Nils98Ar:
Thanks for the update!

mxmxchere:
As it seems there is not really an option to tell containerd to use an http proxy. However i used an env where HTTP_PROXY was set and used nerdctl in it to start a containert. The image was pulled via the proxy then. So for now i think we can cross containerd off the list.

mxmxchere:
[...]

PR #624, opened 2023-10-17
Preserves (possible) additional docker-daemon settings

mxmxchere:
Issue #418 is about using a proxy server in this repo. One portion of this issues is to set the proxy settings for the docker daemon in /etc/docker/daemon.json. This config file is currently provisioned in case the terraform variable mtu is set. If the mtu is set to "0" (default) and the configuration script detects a standard mtu of 1500 for the connecting interface, the file is disposed completely. In order to add other settings, the file should be preserved even if the mtu is standard 1500


DEiselt (approved):
LGTM

PR #625, opened 2023-10-18
Fix kubeapi cidr restrictions

chess-knight:
It seems that from the 'namespace per cluster' feature #481, this was broken

Before, resources were inside the default namespace, so patching did work. But now we have an error

Error: no matches for Id OpenStackCluster.v1alpha7.infrastructure.cluster.x-k8s.io/testcluster.[noNs]; failed to find unique target for patch OpenStackCluster.v1alpha7.infrastructure.cluster.x-k8s.io/testcluster.[noNs]

```ZUUL_CONFIG
restrict_kubeapi = [ "0.0.0.0/0" ]
```


chess-knight:
> Build failed (e2e-quick-test pipeline). https://zuul.scs.community/t/SCS/buildset/f326c4b262e44829aca7e509a97f9e21
> 
> ❌ [k8s-cluster-api-provider-e2e-quick ](https://zuul.scs.community/t/SCS/build/82665eb9746e446bb36045f02349a032) **FAILURE** in 19m 02s
> 
> Warning:
> 
> ## Custom ENV config
> ```
> restrict_kubeapi = [ "none" ]
> ```

`restrict_kubeapi` **works!**

Nils98Ar:
@chess-knight Is the ns of the default sa set with every create_cluster execution? What happens if we run multiple create_cluster for different clusters in parallel?

chess-knight:
> @chess-knight Is the ns of the default sa set with every create_cluster execution? What happens if we run multiple create_cluster for different clusters in parallel?

@Nils98Ar you are right, I will check that and I will probably come up with a more robust namespace extraction

Nils98Ar:
What about:
```
if kubectl get cluster --all-namespaces -o=custom-columns=NAME:.metadata.namespace --no-headers | grep -q ${CLUSTER_NAME}
then 
  CLUSTER_NAMESPACE=${CLUSTER_NAME}
else 
  CLUSTER_NAMESPACE="default"
fi
```

chess-knight:
Looks good.
I was thinking about this:
```
OPENSTACK_CLUSTER_NAMESPACE=$(kubectl get ns $CLUSTER_NAME -o jsonpath='{.metadata.name}' 2>/dev/null)
...
  namespace: ${OPENSTACK_CLUSTER_NAMESPACE:-default}
```

Nils98Ar:
@chess-knight Looks good as well :)

Nils98Ar (approved):
Looks good and works for me!

PR #633, opened 2023-10-25
Add tf openstack to renovate

DEiselt:


mxmxchere:
We want to make sure that terraform does not upgrade to 1.6 or above. This is made sure via the terraform required_version. Renovate seems to honor that setting so we did not an additional pin

mxmxchere (approved):

PR #639, opened 2023-11-01
Delete "*sets" before k8s pods to ensure we enumerate all of them.

### LINKED ISSUE #631: Delete Pods after Deployments
https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/03015416f4cd4109f3a1c89ebe30b29dbd9958c1/terraform/files/bin/delete_cluster.sh#L27

https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/03015416f4cd4109f3a1c89ebe30b29dbd9958c1/terraform/files/bin/delete_cluster.sh#L51

Hi there,

I had a broken cluster and triwed to delete it. Since two of my nodes where not responding (kubelet down) the deleting pods where in terminating state. 
Therefor I had to manually force-delete the not responding pods. Ofc all "deleted" pods were restarting already with the deployments/statefulsets/... sitll in place.

I suggest to switch the lines and delete pods after deleting all other pod-defining things. So move it after "jobs".
Alternatively, since we are deleting all the resources "--force" could also be an approach.

NotTheEvilOne:
This PR just swaps the lines to delete "*sets" before the possibly just recreated k8s pods.

Fixes #631

### DIFF
--- a/terraform/files/bin/delete_cluster.sh
+++ b/terraform/files/bin/delete_cluster.sh
@@ -22,9 +22,6 @@ if echo "$INPODS" | grep nginx >/dev/null 2>&1; then
 	echo -en " Delete ingress \n "
 	timeout 150 kubectl $KCONTEXT delete -f ~/${CLUSTER_NAME}/deployed-manifests.d/nginx-ingress.yaml
 fi
-# Delete pods with persistent volume claims
-echo "Deleting pods with persistent volume claims ..."
-kubectl $KCONTEXT get pods --all-namespaces -o=jsonpath="{'set -x\n'}{range .items[?(@.spec.volumes[*].persistentVolumeClaim.claimName)]}{'kubectl $KCONTEXT delete pod '}{.metadata.name}{' -n '}{.metadata.namespace}{' --grace-period=0 --ignore-not-found=true\n'}{end}" | bash
 # Delete deployments with persistent volume claims
 echo "Deleting deployments with persistent volume claims ..."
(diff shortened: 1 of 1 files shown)


terraform/files/bin/delete_cluster.sh
```diff
@@ -40,6 +37,9 @@ kubectl $KCONTEXT get cronjobs --all-namespaces -o=jsonpath="{'set -x\n'}{range
 # Delete all Jobs
 echo "Deleting all Jobs ..."
 kubectl $KCONTEXT get jobs --all-namespaces -o=jsonpath="{'set -x\n'}{range .items[*]}{'kubectl $KCONTEXT delete job '}{.metadata.name}{' -n '}{.metadata.namespace}{' --grace-period=0 --force\n'}{end}" | bash
+# Delete pods with persistent volume claims
+echo "Deleting pods with persistent volume claims ..."
+kubectl $KCONTEXT get pods --all-namespaces -o=jsonpath="{'set -x\n'}{range .items[?(@.spec.volumes[*].persistentVolumeClaim.claimName)]}{'kubectl $KCONTEXT delete pod '}{.metadata.name}{' -n '}{.metadata.namespace}{' --grace-period=0 --ignore-not-found=true\n'}{end}" | bash
```
chess-knight, line 42: And what about `pods` on line 51? Won't they also be recreated by existing resources below(deployment,daemonset,statefulset)?

NotTheEvilOne: Most likely. Moved these lines as well :)

fdobrovolny:
:thinking: my thinking when I wrote this is that I had a feeling that it may speed up everything as yes they might get recreated but I assume because they get deleted not ready yet it can be faster then properly shutting it down? However it was just my feeling and not based on any research.

chess-knight (approved):
LGTM

PR #645, opened 2023-11-15
Proxy settings for controlplane and worker nodes

mxmxchere:
see #418


mxmxchere:
i was able to execute sonobuoy-quick with http_proxy set. CI will test without http_proxy

DEiselt (approved):
LGTM

PR #651, opened 2023-11-22
New config-setting no_proxy

mxmxchere:
This addresses Issue #418 and introduces a new config variable
name no_proxy. It is trickled down in the same corners where http_proxy
is (when possible). A note to reviewers: When no "no_proxy" variable is
set (from terraforms side). The variable "no_proxy" ends with a trailing
"," (Because i did not spend the effort of adding another case for an empty
"no_proxy" variable in terraform). But this seems to work. Neither could
i notice that a program was confused by the trailing slash nor did any
program treat this as "bypass proxy for all requests".

### DIFF
--- a/terraform/mgmtcluster.tf
+++ b/terraform/mgmtcluster.tf
@@ -98,7 +98,7 @@ write_files:
         "proxies": {
           "http-proxy": "${var.http_proxy}",
           "https-proxy": "${var.http_proxy}",
-          "no-proxy": "127.0.0.0/8,172.18.0.0/16,fc00:f853:ccd:e793::/64,10.96.0.0/16,10.244.0.0/16,kind-control-plane,.svc,.svc.cluster,.svc.cluster.local"
+          "no-proxy": ".svc,.svc.cluster,.svc.cluster.local,127.0.0.0/8,169.254.169.254/32,fd00:ec2::254/128,${var.node_cidr},${var.pod_cidr},${var.service_cidr},${var.no_proxy}"
         },
         %{endif}
         "mtu": ${var.kind_mtu}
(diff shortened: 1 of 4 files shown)


mxmxchere:
I tested with http_proxy set and no_proxy unset. check-quick passed.
I tested with http_proxy set and no_proxy set. check-quick passed.
CI will test with http_proxy and no_proxy unset.

terraform/mgmtcluster.tf
```diff
@@ -119,8 +119,8 @@ write_files:
         export HTTPS_PROXY=$PROXY
         export http_proxy=$PROXY
         export https_proxy=$PROXY
-        export NO_PROXY=172.18.0.0/16,fc00:f853:ccd:e793::/64,10.96.0.0/16,10.244.0.0/16,kind-control-plane,.svc,.svc.cluster,.svc.cluster.local
-        export no_proxy=172.18.0.0/16,fc00:f853:ccd:e793::/64,10.96.0.0/16,10.244.0.0/16,kind-control-plane,.svc,.svc.cluster,.svc.cluster.local
+        export NO_PROXY=".svc,.svc.cluster,.svc.cluster.local,127.0.0.0/8,169.254.169.254/32,fd00:ec2::254/128,${var.node_cidr},${var.pod_cidr},${var.service_cidr},${var.no_proxy}"
+        export no_proxy=".svc,.svc.cluster,.svc.cluster.local,127.0.0.0/8,169.254.169.254/32,fd00:ec2::254/128,${var.node_cidr},${var.pod_cidr},${var.service_cidr},${var.no_proxy}"
```
NotTheEvilOne, line 123: Beside being done that way before, is there any reason why we export environment variables both upper case (which matches convention) and lower case additionally?

NotTheEvilOne (approved):
Looks good for me in general. I just wonder if a variable naming like `proxy_exception_list` would be more suitable. `no_proxy` matches default OpenStack expectation thought [1] so it's fine with me.

[1] https://docs.openstack.org/project-deploy-guide/openstack-ansible/ocata/app-limited-connectivity.html#basic-proxy-configuration

As far as I know `http_proxy` and `https_proxy` are set because of non compliant applications expecting lower case environment variables. Based on documentation it's true for OpenStack as well.

terraform/mgmtcluster.tf
mxmxchere: These env variables are set to be recognized by as many applications as possible. The best way to achieve this (at the cost of potentially unnecessary env variable) is to set both variants.

PR #654, opened 2023-11-30
Add kubeapi mem/cpu limits to the R5 upgrade script

chess-knight:
#552 was missing from R5 upgrade script/cluster-template

I find it during the upgrade process of the SCS container registry cluster SovereignCloudStack/k8s-harbor#54


doc/Upgrade-Guide.md
```diff
@@ -10,8 +10,9 @@ state: Draft (v0.7)
 This document explains the steps to upgrade the SCS Kubernetes cluster-API
 based cluster management solution as follows:
 - from the R2 (2022-03) to the R3 (2022-09) state
-- from the R3 (2022-09) to the R4 state
-- from the R4 (2023-09) to the R5 state
+- from the R3 (2022-09) to the R4 (2023-03) state
+- from the R4 (2023-03) to the R5 (2022-09) state
```
matofeder, line 14: The question is how to port this fix to the v6.0.0. 
IMO I see  several options: 
1. instruct users to cherry-pick this fix commit in the doc
2. release this as v6.0.1 - in this case, it could be tricky to move this commit to the right place and then tag it with v6.0.1

anything else?

matofeder (approved):
LGTM

/tested and it works as expected, thanks @chess-knight for pointing this out!
I left just one comment in the docs part

doc/Upgrade-Guide.md
chess-knight: This document mentions only `maintained/vY.x` branches, so I think the best way is to merge this to the main and backport it to the maintained branches. See 34ebbf10e2d1aba8ee17e07a5945fd0676ebeb54

matofeder: sounds good :+1:

PR #674, opened 2024-01-15
Try k8s v1.29.x

chess-knight:
```ZUUL_CONFIG
kubernetes_version = "v1.29.x"
```


chess-knight:
According to https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/674#issuecomment-1893534588 and also my local testing, only one test failed: `[It] [sig-network] Services should serve endpoints on the same port and different protocols [Conformance]`
This is a new conformance test for v1.29 and according to https://github.com/kubernetes/kubernetes/pull/120069 I found that cilium has a problem with it and that's why the test fails, see https://github.com/cilium/cilium/issues/29913.

chess-knight:
Close, as we add v1.29 to the tech preview list in #673.

We have only one failed conformance test, see https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/674#issuecomment-1893609260 and CAPO and OCCM have no official support for 1.29 yet.

PR #694, opened 2024-02-20
Do not alter clusterclass templates when there is no proxy setting

chess-knight:
Related to #418


matofeder (approved):
Not tested, but it looks good. In my opinion, it's a nice solution on how to get rid of unnecessary cluster template patches.

PR #695, opened 2024-02-21
Update versions in default environment and renovate

chess-knight:
Update versions and renovate in the default environment before R6


matofeder (approved):
lgtm

PR #698, opened 2024-02-21
Update github actions and zuul dependencies

chess-knight:
Update github actions and zuul dependencies before R6


matofeder (approved):
lgtm

PR #700, opened 2024-02-22
Update default cert-manager version to v1.14.2

chess-knight:
Update default cert-manager version to v1.14.2 before R6

```ZUUL_CONFIG
deploy_cert_manager = "true"
```


paulphys (approved):
LGTM

PR #711, opened 2024-03-01
Upgrade R5 to R6

chess-knight:


paulphys (approved):
No objections on my part. Upgrade and clusterclass migration scripts ran through errorless and all modified resources seem to be in order. Tested on fresh cluster based on main branch

PR #714, opened 2024-03-04
Add scs-0210-v2 and scs-0214-v1 zuul checks

matofeder:
Closes https://github.com/SovereignCloudStack/k8s-cluster-api-provider/issues/715


playbooks/tasks/scs_compliance.yaml
```diff
@@ -0,0 +1,53 @@
...
+    ansible.builtin.copy:
+      src: "{{ check_dir }}/Tests/kaas/k8s-version-policy/config.yaml.template"
+      dest: "{{ check_dir }}/Tests/config.yaml"
+      remote_src: true
+      mode: 0644
+  - name: Execute SCS KaaS compliance check
+    ansible.builtin.shell:
+      cmd:
+        ". {{ python_venv_dir }}/bin/activate &&
+         python3 {{ check_dir }}/Tests/scs-compliance-check.py {{ check_dir }}/Tests/scs-compatible-kaas.yaml kaas  -v -s KaaS_V1 -a kubeconfig={{ kubeconfig_path }}"
```
mbuechse, line 43:
```suggestion
         python3 {{ check_dir }}/Tests/scs-compliance-check.py {{ check_dir }}/Tests/scs-compatible-kaas.yaml  -v -s KaaS_V1 -a kubeconfig={{ kubeconfig_path }}"
```

playbooks/tasks/scs_compliance.yaml
```diff
@@ -0,0 +1,53 @@
...
+      virtualenv: "{{ python_venv_dir }}"
+  # TODO: We are simply copying a random check config template here.
+  #  A better solution would be to provide a configuration file that will serve as the source for all SCS checks.
+  #  This issue should be addressed in the scs/standard repository.
+  - name: Copy SCS KaaS compliance config files
+    ansible.builtin.copy:
+      src: "{{ check_dir }}/Tests/kaas/k8s-version-policy/config.yaml.template"
+      dest: "{{ check_dir }}/Tests/config.yaml"
+      remote_src: true
+      mode: 0644
```
mbuechse, line 38: This task can (and should) be dropped.

mbuechse (approved):
Looking good. Thanks very much!

playbooks/tasks/scs_compliance.yaml
matofeder: Then I believe that the new version of scripts won't complain that they cannot find the ./config.yaml. Is that true?

mbuechse: Yes, that is true! We are still working on the [pull request](https://github.com/SovereignCloudStack/standards/pull/508), but so much is clear already.

matofeder: 99c00d842db998107e329187d0c50ea236b3b7fb

mbuechse:
@matofeder: I think with https://github.com/SovereignCloudStack/standards/pull/528 and https://github.com/SovereignCloudStack/standards/pull/508 in place, the newly extended e2e pipeline should now in principle run. What do you think?

matofeder:
@mbuechse Unfortunately, I'm not able to test this e2e pipeline due to GX-SCS infrastructure issues. I've tried to execute the e2e test multiple times, but it always ends with this or a similar error: `Error from server: etcdserver: request timed out.` 

The error is not related to the SCS standard test but to the GX-SCS infrastructure issues and high disk latency.

mbuechse:
@matofeder thanks for the update as well as keeping at it! Much appreciated!

mbuechse:
@matofeder The pipeline ran through fine, except the test scripts produced errors.

- the version policy script seems to have a programming error; we are investigating (@martinmo)
- the node distribution script doesn't seem to see the labels on the nodes, do you by any chance know why this could be? (maybe @cah-hbaum as well)

matofeder:
@mbuechse @cah-hbaum 

Ah, it seems that we have finally a luck to see the SCS standard scripts in action within the end-to-end pipeline. Nice

> the node distribution script doesn't seem to see the labels on the nodes, do you by any chance know why this could be? (maybe @cah-hbaum as well)

IMO both script and pipeline work fine, the issue is that (by default) the e2e pipeline spins up Kubernetes cluster with just one control plane node, i.e. this cluster is not SCS compliant :), Fixed: 56a51141376d83733fa4a95ae42b61c78354f516

mbuechse:
@matofeder Wonderful! The defect in the other script should now also be gone. Shall we say recheck?

matofeder:
> @matofeder Wonderful! The defect in the other script should now also be gone. Shall we say recheck?

It is not needed at the moment. The task to clone the standard repo hasn't been reached in the current pipeline. So, let's wait.

matofeder:
@mbuechse 

See https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/714#issuecomment-2009779389

I have to remove my debug task (which caused the error) and clean up the .zuul.yaml file, but besides these let's call it a success.

chess-knight (approved):
LGTM

I am just wondering, why we do not see Sonobouy results from the last failure https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/714#issuecomment-2012925018

matofeder:
> LGTM
> 
> I am just wondering, why we do not see Sonobouy results from the last failure [#714 (comment)](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/714#issuecomment-2012925018)

Good point, this is because I moved `Parse sonobouy results` from the always block, 
I will try to introduce `always` block in the sonobuoy.yaml as well as in the scs_compliance.yaml and I will move the "parse" tasks and "Insert ... results to the warning message..." tasks there.

PR #718, opened 2024-03-12
Add generation counter for the OpenStackClusterTemplate and fix externalNetworkId

chess-knight:


matofeder (approved):
lgtm

terraform/files/bin/update-R5-to-R6.sh
```diff
@@ -27,6 +27,7 @@ restore() {
   echo "Updating failed ($1)" 1>&2
   cp -p clusterctl.yaml.backup clusterctl.yaml
   cp -p cluster-template.yaml.backup cluster-template.yaml
+  kubectl --context kind-kind patch ValidatingWebhookConfiguration/capo-validating-webhook-configuration --type='json' -p='[{"op": "replace", "path": "/webhooks/0/rules/0/operations", "value":["CREATE", "UPDATE"]}]'
```
matofeder, line 30: Consider the scenario when the user in R5 uses the `external` TF variable and provides an external network ID.
Could be tricky to find out whether the OPENSTACK_EXTERNAL_NETWORK_ID is a network name or ID passed by the user, but in that case we do not need this hack.

matofeder: Already covered in line 79. I overlooked it!

PR #728, opened 2024-04-05
Escape curly braces

maxwolfs:
https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3#bad-usage-of-

### DIFF
--- a/doc/usage/custom-ca.md
+++ b/doc/usage/custom-ca.md
@@ -31,7 +31,7 @@ Steps of what happens with the custom cacert in k8s-cluster-api-provider:
    directory `~/cluster-defaults/${cloud_provider}-cacert` by OpenTofu
 3. During the management server bootstrap process cacert is injected to
    the *~/cluster-defaults/cluster-template.yaml* to *KubeadmControlPlane* and *KubeadmConfigTemplate* files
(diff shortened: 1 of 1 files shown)


doc/usage/custom-ca.md
```diff
@@ -31,7 +31,7 @@ Steps of what happens with the custom cacert in k8s-cluster-api-provider:
    directory `~/cluster-defaults/${cloud_provider}-cacert` by OpenTofu
 3. During the management server bootstrap process cacert is injected to
    the *~/cluster-defaults/cluster-template.yaml* to *KubeadmControlPlane* and *KubeadmConfigTemplate* files
```
mbuechse, line 33:
```suggestion
   the `~/cluster-defaults/cluster-template.yaml` to *KubeadmControlPlane* and *KubeadmConfigTemplate* files
```

mbuechse: I can't suggest these changes everywhere because Github seems to permit this only within the vicinity of your changes, but I would propose to apply similar changes throughout, so we don't switch between emphasis and code for the same purpose.

mbuechse (approved):

PR #738, opened 2024-05-23
Bump default k8s version to v1.28.10

chess-knight:
Resolve `ERROR: The K8s cluster version 1.28.8 of cluster 'main-be9873f-admin@main-be9873f' is outdated according to the standard.`


matofeder (approved):
lgtm

